PR #3, opened 2021-10-18
Nn 3689 create the new adjudications service

richardpopple:


.circleci/config.yml
```diff
@@ -7,11 +7,11 @@ orbs:
 parameters:
   alerts-slack-channel:
     type: string
-    default: hmpps_typescript_notifications
+    default: dps_core_team
 
   releases-slack-channel:
     type: string
-    default: hmpps_typescript_notifications
+    default: dps_core_team
```
richardpopple, line 14: This is where restricted patients goes to so I figured I'd use it here too

PR #21, opened 2021-11-03
NN-3658: Incident details page

natclamp-moj:
Adds the incident details page and connects to the existing statement page, making corrections where necessary


server/data/PrisonLocationResult.ts
```diff
@@ -0,0 +1,24 @@
...
+  agencyId: string
+  parentLocationId: number
+  currentOccupancy: number
+  locationPrefix: string
+  operationalCapacity: number
+  userDescription: string
+  internalLocationCode: string
+}
+
+export type PrisonLocation = Location
```
sp-watson, line 22: Oh - is this a TS feature I don't know? Why not just define these as types?

server/routes/incidentDetails/index.ts
```diff
@@ -0,0 +1,27 @@
+import express, { RequestHandler, Router } from 'express'
+import asyncMiddleware from '../../middleware/asyncMiddleware'
+
+import IncidentDetailsRoutes from './incidentDetails'
+
+import PlaceOnReportService from '../../services/placeOnReportService'
+import LocationService from '../../services/locationService'
+
+export default function prisonerIncidentStatementsRoutes({
```
sp-watson, line 9: Minor naming issue - details not statement

sp-watson (dismissed):

server/routes/incidentDetails/index.ts
natclamp-moj: good catch!

server/data/PrisonLocationResult.ts
natclamp-moj: Good point, something that came with merging a couple of files but i'll simplify

sp-watson (approved):

PR #25, opened 2021-11-04
NN-3668: Initial page and backend code for placed on report confirmation page

sp-watson:


server/routes/confirmedOnReport/confirmedOnReport.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export default class ConfirmedOnReportRoutes {
+  constructor(private readonly reportedAdjudicationsService: ReportedAdjudicationsService) {}
+
+  private renderView = async (req: Request, res: Response): Promise<void> => {
+    const { adjudicationNumber } = req.query
+    const { user } = res.locals
+
+    const adjudicationNumberValue: number = parseInt(adjudicationNumber as string, 10)
+    if (Number.isNaN(adjudicationNumberValue)) {
```
richardpopple, line 14: Out of interest I guess this results in a 500? But I guess that's okay as we should always have an adjudication number. Do our users mess with urls?

server/utils/utils.ts
```diff
@@ -21,6 +21,9 @@ export const isBlank = (str: string): boolean => !str || /^\s*$/.test(str)
  */
 export const properCaseName = (name: string): string => (isBlank(name) ? '' : name.split('-').map(properCase).join('-'))
 
+export const formatName = (firstName: string, lastName: string): string =>
```
richardpopple, line 24: Ah - I will use this in my next pull request - cheers

richardpopple (approved):

server/routes/confirmedOnReport/confirmedOnReport.ts
sp-watson: Actually that's a good point - it should be a 400. I will sort that out in a later PR

richardpopple (approved):

PR #35, opened 2021-11-10
NN-3666: Add reverse for surname in reporting officer string

natclamp-moj:


server/utils/utils.ts
```diff
@@ -27,8 +27,11 @@ export const formatName = (firstName: string, lastName: string): string =>
   [properCaseName(firstName), properCaseName(lastName)].filter(Boolean).join(' ')
 
 export const getFormattedReporterName = (name: string): string => {
+  if (!name) {
+    return null
+  }
   const names = name.split(' ')
-  return `${names[0][0]}. ${names[1]}`
+  return `${names[0][0]}. ${names.reverse()[0]}`
```
sp-watson, line 34: Isn't the first and last name already known when this is called? If it is then I guess it would be better to pass those in rather than the formatted name

sp-watson (approved):

PR #42, opened 2021-11-12
NN-3723 your completed reports page

richardpopple:


server/data/manageAdjudicationsClient.ts
```diff
@@ -52,4 +57,23 @@ export default class ManageAdjudicationsClient {
...
+
+  async getCompletedAdjudications(
+    user: User,
+    pageRequest: PageRequest
+  ): Promise<PageResponse<CompletedAdjudicationSummary>> {
+    return pageResponseFrom(pageRequest, dummyResults)
+  }
 }
+
+const dummyResults: CompletedAdjudicationSummary[] = generateRange(1, 1, _ => {
```
JamieDraperUK, line 69: Should this live in a test helper module of some sort?

server/routes/yourCompletedReports/index.ts
```diff
@@ -0,0 +1,22 @@
+import express, { RequestHandler, Router } from 'express'
+import asyncMiddleware from '../../middleware/asyncMiddleware'
+
+import YourCompletedReports from './yourCompletedReports'
+
+import CompletedAdjudicationsService from '../../services/completedAdjudicationsService'
+
+export default function yourCompletedReportsRoutes({
```
JamieDraperUK, line 8: Just checking... Is the casing of the name correct?

server/data/manageAdjudicationsClient.ts
richardpopple: Yeah - I can pull it out somewhere else - its basically to be purged when I'm done - the only reason its here is so that I can test what the front end looks like before the back end apis are in place

server/routes/yourCompletedReports/yourCompletedReports.ts
```diff
@@ -0,0 +1,30 @@
...
+  private renderView = async (
+    req: Request,
+    res: Response,
+    results: PageResponse<CompletedAdjudicationSummary>
+  ): Promise<void> =>
+    res.render(`pages/yourCompletedReports`, {
+      yourCompletedReports: results,
+      pagination: mojPaginationFromPageResponse(
+        results,
+        new URL(`${req.protocol}://${req.get('host')}${req.originalUrl}`)
```
JamieDraperUK, line 19: I'm probably wrong here but worth checking.. doesn't `req.originalUrl `already give you the protocol and host?

server/routes/yourCompletedReports/yourCompletedReports.ts
```diff
@@ -0,0 +1,30 @@
...
+      ),
+    })
+
+  view = async (req: Request, res: Response): Promise<void> => {
+    const results = await this.completedAdjudicationsService.getCompletedAdjudications(
+      res.locals.user,
+      pageRequestFrom(20, +req.query.pageNumber || 1)
```
JamieDraperUK, line 26: Should there be a plus before `+req.query.pageNumber` I've not seen that before

JamieDraperUK (approved):
LGTM

server/routes/yourCompletedReports/yourCompletedReports.ts
richardpopple: Yeah - I don't know if this is considered nasty or not - but it converts the string into a number - which typescript was complaining about

server/routes/yourCompletedReports/yourCompletedReports.ts
richardpopple: I was hoping that was the case - but it doesn't seem to - I will investigate further though

server/routes/yourCompletedReports/index.ts
richardpopple: You mean instead of YourCompletedReportsRoutes? I think that lowercase is the norm here - or at least it seems to be with the other roots.

PR #55, opened 2021-11-18
NN-3723 completed reports

richardpopple:


server/services/completedAdjudicationsService.ts
```diff
@@ -1,16 +1,58 @@
...
+  ): ReportedAdjudicationEnhanced {
+    const displayName =
+      (prisonerResult && convertToTitleCase(`${prisonerResult.lastName}, ${prisonerResult.firstName}`)) || ''
+    const friendlyName =
+      (prisonerResult && convertToTitleCase(`${prisonerResult.firstName} ${prisonerResult.lastName}`)) || ''
+    return {
+      ...reportedAdjudication,
+      displayName,
+      friendlyName,
+      dateTimeOfIncident: timestampToDate(reportedAdjudication.incidentDetails.dateTimeOfIncident),
```
sp-watson, line 51: I am not sure there is any need to use this new function to convert to a date, as it is eventually used for sorting.... and i don't think the sorting code (in JS) knows what a Date is

I don't know - it might work (maybe you took this from DPS?) - just checking

server/routes/yourCompletedReports/yourCompletedReports.ts
```diff
@@ -1,27 +1,27 @@
...
     req: Request,
     res: Response,
-    results: PageResponse<CompletedAdjudicationSummary>
+    results: PageResponse<ReportedAdjudication>
   ): Promise<void> =>
     res.render(`pages/yourCompletedReports`, {
       yourCompletedReports: results,
       pagination: mojPaginationFromPageResponse(
-        results,
+        results.changeIndex(1),
```
sp-watson, line 18: It took me a while to figure out what this was doing.

I think we can review the pagination abstraction to make this a bit clearer.

Something for later though (not as important as doing other tickets).

sp-watson (approved):

server/services/completedAdjudicationsService.ts
richardpopple: So - you are correct - its for sorting. The sorting seems to work as expected in the front end

PR #71, opened 2021-11-25
NN-3764: Fix bug in time validation

natclamp-moj:


server/routes/incidentDetails/incidentDetailsValidation.ts
```diff
@@ -80,11 +81,15 @@ export default function validateForm({
   if (incidentDate.time.minute.length < 2) {
     return errors.ONE_DIGIT_MINUTE
   }
-  if (incidentDate.date === `${now.getDate()}/${now.getMonth() + 1}/${now.getFullYear()}`) {
-    if (Number(incidentDate.time.hour) > now.getHours()) {
+
+  if (incidentDate.date === getDate(now, 'DD/MM/YYYY')) {
```
sp-watson, line 85: Surely you don't need to strip the hours and minutes?!

Can'y you just get the UTC timestamp (`new Date().getTime()`) and diff that with the provided time?

It will require converting the provided hours and minutes to a UTC, but that should be pretty easy.

Worth investigating...

natclamp-moj: Yeah you're right, must have been 6pm delirium

sp-watson (approved):

PR #76, opened 2021-12-02
NN-3729: Add not found redirect

natclamp-moj:


server/routes/allCompletedReports/allCompletedReports.ts
```diff
@@ -21,6 +26,10 @@ export default class AllCompletedReportsRoutes {
     })
 
   view = async (req: Request, res: Response): Promise<void> => {
+    const userRoles = await this.userService.getUserRoles(res.locals.user.token)
```
richardpopple, line 29: I take it we show a not found page instead of an access denied one?

richardpopple (approved):

PR #78, opened 2021-12-03
NN-3776: Prisoner report

natclamp-moj:


server/routes/index.ts
```diff
@@ -30,6 +31,11 @@ export default function routes(
   router.use('/your-completed-reports', yourCompletedReportsRoutes({ reportedAdjudicationsService }))
   router.use('/all-completed-reports', allCompletedReportsRoutes({ reportedAdjudicationsService, userService }))
   router.use('/select-report', continueReportSelectRoutes({ placeOnReportService }))
+  router.use('/all-completed-reports', allCompletedReportsRoutes({ reportedAdjudicationsService, userService }))
```
sp-watson, line 34: Duplicate?!

server/services/reportedAdjudicationsService.ts
```diff
@@ -124,4 +133,48 @@ export default class ReportedAdjudicationsService {
...
+    const { draftAdjudication } = newDraftAdjudicationData
+    const reporter = await this.hmppsAuthClient.getUserFromUsername(draftAdjudication.createdByUserId, user.token)
+
+    const dateTime = draftAdjudication.incidentDetails.dateTimeOfIncident
+    const date = getDate(dateTime, 'D MMMM YYYY')
+    const time = getTime(dateTime)
+
+    const [locationObj] = locations.filter(loc => loc.locationId === draftAdjudication.incidentDetails.locationId)
+
+    const incidentDetails = [
```
sp-watson, line 154: Hmm.. this appears to be UI code, so should really be in the njk somehow

server/routes/prisonerReport/index.ts
```diff
@@ -0,0 +1,28 @@
...
+  placeOnReportService: PlaceOnReportService
+  locationService: LocationService
+}): Router {
+  const router = express.Router()
+
+  const prisonerReport = new PrisonerReportRoutes(reportedAdjudicationsService, placeOnReportService, locationService)
+
+  const get = (path: string, handler: RequestHandler) => router.get(path, asyncMiddleware(handler))
+
+  get('/:prisonerNumber/:adjudicationNumber', prisonerReport.view)
```
richardpopple, line 25: I assume that we actually only logically need the adjudication number at the prisonerNumber is just there to make things nicer in the url

server/views/macros/printLink.njk
```diff
@@ -0,0 +1,7 @@
+{% macro printLink(linkText = 'Print this page', align = "left", classes = "") %}
```
sp-watson, line 1: Don't we already have this? I thought we made a macro to show the print link?

server/views/pages/prisonerReport.njk
```diff
@@ -0,0 +1,51 @@
...
+ <div class="prisonerPlaybackSection">
+  {{ prisonerPlaybackSection(prisoner.displayName, prisoner.prisonerNumber, prisoner.currentLocation) }}
+ </div>
+
+ <div class="prisoner-report-top-bar govuk-!-margin-top-5">
+  <p class="govuk-body"><span class="govuk-!-font-weight-bold">Report number: </span>{{ data.reportNo }}</p>
+  {{ printLink('Print a copy of this report', 'right', 'govuk-!-padding-right-0') }}
+ </div>
+
+ <div class="govuk-!-margin-top-5">
```
sp-watson, line 31: How similar is all this to `checkYourAnswers.njk`? Is it a strong candidate for a macro?

server/utils/utils.ts
```diff
@@ -105,6 +105,16 @@ export const possessive = (string: string): string => {
   return `${string}${string.toLowerCase().endsWith('s') ? '’' : '’s'}`
 }
 
+export const formatStatement = (statement: string): string => {
```
sp-watson, line 108: I know it was already in the ts code, but this really belongs in the njk somehow.

server/routes/prisonerReport/index.ts
natclamp-moj: I use the prisonerNumber from the params in the controller as we need to call prisonAPI to get the prisoner details for the header thingy (location, name etc)

server/services/reportedAdjudicationsService.ts
```diff
@@ -124,4 +133,48 @@ export default class ReportedAdjudicationsService {
       ),
     }
   }
+
+  async getPrisonerReport(
+    user: User,
+    adjudicationNumber: number,
+    locations: PrisonLocation[]
+  ): Promise<PrisonerReport> {
+    const newDraftAdjudicationData = await new ManageAdjudicationsClient(
```
richardpopple, line 142: What is the correct behaviour here if we hit this url twice? Does createDraftFromCompleteAdjudication simply return the adjudication that had been recreated the first time do you know?

richardpopple: Or do we need to check that one doesn't already exist?

natclamp-moj: This is one of the things I wanted to check with you both. It just creates a new one every time - I thought about adding a check in there, but we have no information to check with? We don't have the new draft id that was created the first time we hit it...

server/routes/index.ts
natclamp-moj: Good spot, not sure how that got there

server/views/pages/prisonerReport.njk
natclamp-moj: For some reason Eddie has changed where the 'change' links are so some of the CSS is quite different

server/views/macros/printLink.njk
natclamp-moj: Seems not. I looked for it as I thought I'd added it previously, but seems not.

server/views/pages/prisonerReport.njk
natclamp-moj: But i was going to macro some of it for the other prisoner-report page when I started that

server/utils/utils.ts
natclamp-moj: Have moved this to the nunjucks setup file as a filter

server/views/pages/prisonerReport.njk
sp-watson: Oh yeah - surely that's an error on his behalf? Can we check, because it seems unnecessary to have 2 versions of what is effectively the same data just for Change link positioning?!

If he doesn't agree, we'll make hime!

server/services/reportedAdjudicationsService.ts
sp-watson: To quote my ticket for the work https://dsdmoj.atlassian.net/browse/NN-3780:

"ATM this endpoint will not make any effort to prevent duplicate drafts of the same reported adjudication".

It is the responsibility of the API to do the check, but it gets really complicated when you consider they might (somehow) have made changes to the existing draft.

server/routes/prisonerReport/index.ts
sp-watson: Yeah, i am not sure about that, because we have to get the adjudication anyway, and that contains the prisoner number.

But I don't mind that too much ATM

sp-watson (approved):

PR #90, opened 2021-12-10
NN-3850: Confirmation page for reporter when they edit the report

natclamp-moj:


server/services/reportedAdjudicationsService.ts
```diff
@@ -47,6 +47,22 @@ export default class ReportedAdjudicationsService {
     }
   }
 
+  async getSimplifiedReportedAdjudication(adjudicationNumber: number, user: User): Promise<ConfirmedOnReportData> {
```
sp-watson, line 50: I really don't like this name - it would be better to call it something like `getConfirmationDetails` and maybe rename the other one to indicate it returns other information for further instructions - `getDetailsForInstruction` (I can't think of a good one right now)

There is no need to add ReportedAdjudication as that is part of the service name

server/views/pages/confirmedOnReport.njk
```diff
@@ -6,35 +6,51 @@
...
+      {{ govukPanel({
+          titleText: bannerText,
+          html: bannerHTML,
+          classes: "govuk-!-margin-bottom-6"
+        }) }}
+
+      <h2 class="govuk-heading-m">What you must do next</h2>
+
+      <div class="confirmed-on-report-section">
+        {% if not changed %}
```
sp-watson, line 30: Could we have a more descriptive name - this name indicates something has changed, but what. Maybe try `adjudicationEdited`?

sp-watson (approved):

PR #96, opened 2021-12-14
NN-3854: Check your answers and confirmation pages (reviewer)

natclamp-moj:


sp-watson:
The name of this PR is important, as it is what we all see in the history...

It should have the format NN-xxxx: ....

server/routes/incidentDetails/incidentDetailsSubmittedEdit.ts
```diff
@@ -51,13 +52,18 @@ export default class IncidentDetailsSubmittedEditRoutes {
       locationId: location,
     }
 
+    const cancelButtonHref = (referrer as string)?.includes('review')
```
sp-watson, line 55: Best practice is to avoid relying on a URL name (or any string that might change) to decide what to do next.

But there is no easy fix with the way things are right now.. I'll make a note

server/routes/prisonerReport/prisonerReportReview.ts
```diff
@@ -35,7 +35,7 @@ export default class prisonerReportReviewRoutes {
       prisoner,
       data,
       printHref: `/print-report/${adjudicationNumber}?referrer=/prisoner-report/${prisoner.prisonerNumber}/${adjudicationNumber}/review`,
-      editIncidentDetailsURL: `/incident-details/${prisoner.prisonerNumber}/${data.draftId}/submitted/edit`,
+      editIncidentDetailsURL: `/incident-details/${prisoner.prisonerNumber}/${data.draftId}/submitted/edit?referrer=/prisoner-report/${prisoner.prisonerNumber}/${adjudicationNumber}/review`,
```
sp-watson, line 38: Maybe it's not important, but I would have expected tests to fail - and be changed - because of this?!

server/routes/prisonerReport/prisonerReport.ts
```diff
@@ -31,7 +31,7 @@ export default class prisonerReportRoutes {
       prisoner,
       data,
       printHref: `/print-report/${adjudicationNumber}?referrer=/prisoner-report/${prisoner.prisonerNumber}/${adjudicationNumber}/report`,
-      editIncidentDetailsURL: `/incident-details/${prisoner.prisonerNumber}/${data.draftId}/submitted/edit`,
+      editIncidentDetailsURL: `/incident-details/${prisoner.prisonerNumber}/${data.draftId}/submitted/edit?referrer=/prisoner-report/${prisoner.prisonerNumber}/${adjudicationNumber}/report`,
```
sp-watson, line 34: Maybe it's not important, but I would have expected tests to fail - and be changed - because of this?!

sp-watson (approved):

richardpopple (approved):

PR #97, opened 2021-12-15
NN-3721 print form

richardpopple:


assets/js/main.js
```diff
@@ -1,4 +1,4 @@
-$(document).ready(function () {
+jQuery(function () {
```
richardpopple, line 1: This is apparently the none deprecated way

server/middleware/setUpWebSecurity.ts
```diff
@@ -13,13 +13,7 @@ export default function setUpWebSecurity(): Router {
         directives: {
           defaultSrc: ["'self'"],
           // Hash allows inline script pulled in from https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/template.njk
-          scriptSrc: [
-            "'self'",
-            'code.jquery.com',
-            "'sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU='",
-            "'sha256-loNEcvOiwPZFEqTNJYE+/sdSv0X/6zDcCFyxAG+oGiw='",
```
richardpopple, line 20: Firefox does not support unsafe-hashes for inline event handlers - so I've removed this and added some jquery in main (it supports inline script though - so I've left that bit in (from the gds template))

server/routes/printReport/NoticeOfBeingPutOnReportData.ts
```diff
@@ -0,0 +1,36 @@
+import { ConfirmedOnReportData } from '../../data/ConfirmedOnReportData'
+import { convertToTitleCase, formatTimestampTo } from '../../utils/utils'
+
+export default class NoticeOfBeingPutOnReportData {
```
richardpopple, line 4: I've used an object to pass about the template because there were a lot of properties and it was easy to get the ordering wrong, and its used in 2 places so generating the properties in one place made sense

server/routes/printReport/NoticeOfBeingPutOnReportData.ts
```diff
@@ -0,0 +1,36 @@
+import { ConfirmedOnReportData } from '../../data/ConfirmedOnReportData'
```
sp-watson, line 1: Not exciting, but given that this is just a class, maybe better to put in the `data` directory.

We can do that later though

sp-watson:
A lot of this print stuff is way over my head!

sp-watson (approved):

PR #98, opened 2021-12-15
NN-3721: print form

richardpopple:


assets/sass/local.sass
```diff
@@ -17,7 +17,3 @@ body
 
   .govuk-main-wrapper
     padding: 0px !important /* Remove the initial padding on the first page */
-
```
richardpopple, line 20: Dom is seeing this weird issue where his version of edge fails to load the screen - but magically if he specifies the border type in the print dialog the issue goes away - this is the only thing I can think that might have that effect so I'm pulling it out for now and seeing if it fixes his issue. Unfortunately it means we have less control over the rendering and so I've had to change the html so it still fits the pages in firefox

natclamp-moj (approved):

PR #107, opened 2022-01-17
NN-3902: Staff offender search functionality

natclamp-moj:


server/views/pages/associatedStaffSelect.njk
```diff
@@ -0,0 +1,104 @@
...
+    }) }}
+  {% endif %}
+
+  <h1 class="govuk-heading-l">{{ title }}</h1>
+
+  <form method="POST" novalidate="novalidate" data-qa="search-form">
+  <input type="hidden" name="_csrf" value="{{ csrfToken }}" />
+
+  <div class="horizontal-form govuk-!-margin-bottom-3">
+    {{ govukInput({
```
sp-watson, line 30: This has to match the inputs in the `assault` form doesn't it? 

If there are 2 inputs there - first and last name - then there needs to be 2 inputs here

sp-watson (approved):

PR #108, opened 2022-01-17
NN-3895: Cater for API updates to include incident role

sp-watson:
This requires temporary code to set the incident role when creating the adjudication.

The rest of the code is to update the API client to match the API changes, and add to tests


natclamp-moj (approved):
lgtm

PR #110, opened 2022-01-18
NN-3912: Use createdDateTime for print form

natclamp-moj:


server/views/macros/noticeOfBeingPlacedOnReportDetails.njk
```diff
@@ -57,10 +57,10 @@
             officer: {{ noticeOfBeingPlacedOnReportData.reportingOfficer }}</p>
         <p class='govuk-body govuk-!-margin-bottom-0 govuk-!-font-size-14'
            data-qa='notice-of-being-placed-on-report-incident-date'>
```
sp-watson, line 59: This label is a bit confusing - worth sorting out eventually.

ok for now though

sp-watson (approved):

PR #113, opened 2022-01-28
NN-3866: Staff member page refactor

natclamp-moj:


server/routes/selectAssociatedStaff/selectAssociatedStaff.test.ts
```diff
@@ -73,11 +92,11 @@ describe('GET /select-associated-staff', () => {
 describe('POST /select-associated-staff', () => {
   it('should redirect to select staff member page with the correct search text and redirect URL intact', () => {
     return request(app)
-      .post('/select-associated-staff?searchFirstName=john&searchLastName=smith')
-      .send({ searchFirstName: 'john', searchLastName: 'doe' })
+      .post('/select-associated-staff?staffFirstName=john&staffLastName=smith')
```
sp-watson, line 95: A bit confused by this (sorry, didn't spot this last time) - how come the first/last name are in the URL and the post body?

server/routes/selectAssociatedStaff/selectAssociatedStaff.ts
```diff
@@ -1,27 +1,29 @@
...
+import UserService from '../../services/userService'
 import validateForm from './staffSearchValidation'
+import PlaceOnReportService, { StaffSearchByNameEnhanced } from '../../services/placeOnReportService'
 
 type PageData = {
   error?: FormError
-  searchResults?: StaffSearchByName[]
-  searchFirstName: string
-  searchLastName: string
+  searchResults?: StaffSearchByNameEnhanced[]
```
sp-watson, line 10: I would prefer this to have a name that represents what it contains - so `StaffSearchByWithDetails` or something?

server/services/placeOnReportService.ts
```diff
@@ -225,4 +230,17 @@ export default class PlaceOnReportService {
...
+    const token = await this.hmppsAuthClient.getSystemClientToken(user.username)
+
+    const getLocationName = async (staffMember: StaffSearchByName) => {
+      if (staffMember.activeCaseLoadId === 'CADM_I') return { ...staffMember, currentLocation: 'Central Admin' }
+
+      const locationName = await new PrisonApiClient(token).getAgency(staffMember.activeCaseLoadId)
+      return { ...staffMember, currentLocation: locationName?.description }
+    }
+
+    return Promise.all(staffMembers.map((staffMember: StaffSearchByName) => getLocationName(staffMember)))
```
sp-watson, line 244: It's not great to make network calls for every search row, especially if we aren't paginating.

What we tend to do to minimise network calls is to remove duplicates before making the network calls.

server/views/pages/associatedStaffSelect.njk
```diff
@@ -76,9 +76,9 @@
           "data-sort-value": staffMember.name
         }
       },
-      { text: staffMember.activeCaseLoadId },
+      { text: staffMember.currentLocation or staffMember.activeCaseLoadId },
```
sp-watson, line 79: Will currentLocation ever be null?

server/services/placeOnReportService.test.ts
```diff
@@ -582,4 +583,70 @@ describe('placeOnReportService', () => {
...
+          name: 'Tester Person',
+          staffId: 1234564789,
+          username: 'RO_USER_TEST',
+          verified: true,
+        },
+      ]
+      const response = await service.getAssociatedStaffDetails(staffMembers, user)
+      expect(response).toEqual([
+        {
+          activeCaseLoadId: 'MDI',
```
sp-watson, line 611: Just a note for the future (and now if you want) - I think it is better to use the input as a basis for the expected output, if that is what is expected.

In this case, you have almost all these details in `staffMembers` - it is much less typing, and makes it clearer what you are testing, if you just inherit those values.

PR #119, opened 2022-02-04
NN-3935: prototype for offence question data structure

richardpopple:


server/offenceCodeDecisions/Commited.ts
```diff
@@ -0,0 +1,153 @@
+import Decision from './Decision'
+import Question from './Question'
+import Code from './Code'
+
+// Codes
```
richardpopple, line 5: Need to go through the list of codes in the spread sheet and fill in the values. Might need a sanity check when doing this.

server/offenceCodeDecisions/Commited.ts
```diff
@@ -0,0 +1,153 @@
...
+const todo = new Code('TODO')
+
+function decision(question: Question | string) {
+  return new Decision(question)
+}
+
+export default new Decision()
```
richardpopple, line 12: This is the data structure for the decisions that need to be made when choosing the offender committed an offence. There need to be 3 more for the other selection on the first page. They are very similar to each other and it would be possible to refactor them all into one instance I think. However I've not done this because:
1) While they are very similar at the moment there is no logical reason that they would be in the future and if they all in one structure having differences will be tricky
2) Refactoring to make all of the structures fit into one makes the data outlined below harder to read.
So instead we'll have 4 instances like this one which more verbose.
I'm not wedded to either approach if anyone has strong feeling happy to go with that!

server/offenceCodeDecisions/Commited.ts
```diff
@@ -0,0 +1,153 @@
...
+          .title('What happened?')
+          .child(
+            decision('Has unauthorised article in possession')
+              .title('Did they have a greater amount than they are allowed to have?')
+              .child(decision('Yes').code(todo))
+              .child(decision('No').code(todo))
+          )
+          .child(
+            decision('Sells or gives an unauthorised article to another person')
+              .title('Was the article only for {OFFENDED_FIRST_NAME} {OFFENDER_LAST_NAME}`s personal user')
```
richardpopple, line 71: We have to do token replacement in some of the questions - its pretty simple matching

server/offenceCodeDecisions/Decision.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  constructor(question?: Question | string) {
+    if (question instanceof Question) {
+      this.decisionQuestion = question
+    } else if (typeof question === 'string') {
+      this.decisionQuestion = new Question(question)
+    }
+  }
+
+  id(): string {
```
richardpopple, line 28: We generate an id based on the index where the decision lives in the tree.
i.e. 0 is the root
0:0 is the roots first child etc.
As they are calculated like this the ids will change if we add or remove nodes in future - so we shouldn't persist them anywhere I think

server/offenceCodeDecisions/Decision.ts
```diff
@@ -0,0 +1,191 @@
...
+    }
+    return codes.sort()
+  }
+
+  invalidDecisions(): Array<Decision> {
+    return this.matching(d => d.invalid())
+  }
+
+  invalid(): boolean {
+    return (
```
richardpopple, line 146: Invalid if:
We have any leaf nodes that done't have a code
We have node with children that doesn't have a title
The page or url starts with a /. This is just to ensure consistency when using them to build paths in the controllers

server/offenceCodeDecisions/Decision.ts
```diff
@@ -0,0 +1,191 @@
...
+      (this.getChildren().length === 0 && this.getCode() == null) ||
+      (this.getChildren().length !== 0 && this.getTitle() == null) ||
+      (this.getUrl() != null && this.matching(d => d !== this && d.getUrl() === this.getUrl()).length !== 0) ||
+      (this.getUrl() != null && this.getUrl().startsWith('/')) ||
+      (this.getPage() != null && this.getPage().startsWith('/'))
+    )
+  }
+
+  questionsToGetHere(): Array<Question> {
```
richardpopple, line 155: Given a decision node - get the questions that are required to get here. This should allow us to specify on the summary page the questions that we answered hopefully. There is still some nastiness to do with sticking extra data in the session - I reckon that'll be part of question type.

server/offenceCodeDecisions/Decision.ts
```diff
@@ -0,0 +1,191 @@
...
+    let questions = new Array<Question>()
+    if (this.getParent()) {
+      questions = this.getParent().questionsToGetHere()
+    }
+    if (this.getQuestion()) {
+      questions.push(this.getQuestion())
+    }
+    return questions
+  }
+
```
richardpopple, line 165: Given a code find the decision that give it. Currently not very useful because they are all the same, but when they are updated it will allow us to find the unique node and thus the questions that led to it.

server/offenceCodeDecisions/Decision.ts
```diff
@@ -0,0 +1,191 @@
...
+  findByCode(code: Code): Decision {
+    return this.findBy(d => d.getCode()?.code === code?.code)
+  }
+
+  toString(indent = 0): string {
```
richardpopple, line 170: This method should output the decisions in an indented fashion which is useful when building them to check they look right.....

server/offenceCodeDecisions/Decisions.test.ts
```diff
@@ -0,0 +1,43 @@
...
+        return null
+      }
+      return item
+    })
+    .filter(item => item != null)
+}
+
+describe('decisions', () => {
+  it.skip('toString', () => {
+    // Not a test but useful output
```
richardpopple, line 18: This test is skipped - but if you want to see what the structure looks like then comment this back in

server/offenceCodeDecisions/Decisions.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    decisionTrees.forEach(decisionTree => console.log(decisionTree.toString()))
+  })
+
+  it.skip('no duplicate codes', () => {
```
richardpopple, line 22: skipped until the codes have been filled in.

server/offenceCodeDecisions/Decisions.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    decisionTrees.forEach(decisionTree => {
+      const allCodes = decisionTree.allCodes().map(c => c.code)
+      const duplicates = findDuplicates(allCodes)
+      expect(duplicates).toHaveLength(0)
+    })
+  })
+
+  it('no invalid decisions', () => {
+    decisionTrees.forEach(decisionTree => {
+      expect(decisionTree.invalidDecisions().map(d => d.getQuestion())).toHaveLength(0)
```
richardpopple, line 32: Check that the structure is valid.

server/offenceCodeDecisions/Decision.ts
```diff
@@ -0,0 +1,191 @@
...
+  getChildren() {
+    return this.childrenDecisions
+  }
+
+  getParent() {
+    return this.parentDecision
+  }
+
+  getUrl(): string {
+    return this.decisionUrl || this.id()
```
richardpopple, line 90: If we don't explicitly specify a url we use the id

server/offenceCodeDecisions/IncidentRoles.ts
```diff
@@ -0,0 +1,3 @@
+type IncidentRole = 'committed'
```
richardpopple, line 1: Still not worked out how to make enums work. Will need to revisit this later

server/offenceCodeDecisions/Placeholder.ts
```diff
@@ -0,0 +1,12 @@
+export type PlaceholderValues = {
```
richardpopple, line 1: Functionality to fill placeholders in the questions and titles

server/offenceCodeDecisions/Question.ts
```diff
@@ -0,0 +1,17 @@
+import processText, { PlaceholderValues } from './Placeholder'
+
+type QuestionType = 'RADIO' | 'FIND_PRISONER'
```
richardpopple, line 3: again with the enums....

server/routes/index.ts
```diff
@@ -19,13 +19,15 @@ import printReportRoutes from './printReport'
...
 import selectAssociatedStaffRoutes from './selectAssociatedStaff'
+import offenceCodeDecisionsRoutes from './offenceCodeDecisions'
 
 import { Services } from '../services'
 
 export default function routes(
   router: Router,
   { placeOnReportService, locationService, prisonerSearchService, reportedAdjudicationsService, userService }: Services
 ): Router {
+  router.use('/offence-code', offenceCodeDecisionsRoutes({ placeOnReportService }))
```
richardpopple, line 30: Don't know if this is the url to use - I think we'll need to discuss these when we start wiring up journeys

server/routes/offenceCodeDecisions/index.ts
```diff
@@ -0,0 +1,18 @@
...
+export default function offenceCodeDecisionsRoutes({
+  placeOnReportService,
+}: {
+  placeOnReportService: PlaceOnReportService
+}): Router {
+  const router = express.Router()
+  const offenceCodeDecisions = new OffenceCodeDecisionsRoutes(placeOnReportService)
+  const get = (path: string, handler: RequestHandler) => router.get(path, asyncMiddleware(handler))
+  const post = (path: string, handler: RequestHandler) => router.post(path, asyncMiddleware(handler))
+  get('/:adjudicationNumber/:incidentRole/*', offenceCodeDecisions.view)
```
richardpopple, line 15: The star matches either the id of the decision or the url of the decision (the same mechanism us used for both)

server/routes/offenceCodeDecisions/offenceCodeDecisions.ts
```diff
@@ -0,0 +1,84 @@
...
+
+type PageData = {
+  error?: FormError
+  selectedDecisionId?: string
+}
+
+export default class OffenceCodeRoutes {
+  constructor(private readonly placeOnReportService: PlaceOnReportService) {}
+
+  private async placeholderValues(adjudicationNumber: string, res: Response) {
```
richardpopple, line 19: This should maybe go in a service - but I have a suspicion that we'll need to also grab stuff off the session - so maybe a helper even.

server/routes/offenceCodeDecisions/offenceCodeDecisions.ts
```diff
@@ -0,0 +1,84 @@
...
+      offenderFirstName: properCaseName(prisonerDetails.firstName),
+      offenderLastName: properCaseName(prisonerDetails.lastName),
+    }
+  }
+
+  private renderView = async (req: Request, res: Response, pageData?: PageData): Promise<void> => {
+    const { adjudicationNumber, incidentRole } = req.params
+    const { error, selectedDecisionId } = pageData
+    const placeholderValues = await this.placeholderValues(adjudicationNumber, res)
+    const path = req.path.replace(`/${adjudicationNumber}/${incidentRole}/`, '')
```
richardpopple, line 39: we grab everything after the incidentRole (which I've just hard coded for the time being and will need to be consistent with the rest of the application) and use it to find the decision in question

server/routes/offenceCodeDecisions/offenceCodeDecisions.ts
```diff
@@ -0,0 +1,84 @@
...
+    if (error) {
+      return this.renderView(req, res, {
+        error,
+        selectedDecisionId,
+      })
+    }
+
+    const selectedDecision = decisionTrees.get(incidentRole as IncidentRole).findById(selectedDecisionId)
+    const redirectUrl = selectedDecision.getCode()
+      ? `/TODO`
```
richardpopple, line 75: We've got to the end of a decision tree - need to discuss where this then goes too

server/routes/offenceCodeDecisions/offenceCodeDecisionsValidation.ts
```diff
@@ -0,0 +1,21 @@
...
+type DecisionForm = {
+  selectedDecisionId?: string
+}
+
+type ErrorType = 'MISSING_SELECTION'
+
+const errors: { [key in ErrorType]: FormError } = {
+  MISSING_SELECTION: {
+    href: '#selectedDecisionId',
+    text: 'Please make a choice', // TODO correct message
```
richardpopple, line 12: Not sure what this message should be at the moment.

server/routes/offenceCodeDecisions/offenceCodeDecisions.ts
```diff
@@ -0,0 +1,84 @@
...
+    })
+  }
+
+  view = async (req: Request, res: Response): Promise<void> => this.renderView(req, res, {})
+
+  submit = async (req: Request, res: Response): Promise<void> => {
+    const { adjudicationNumber, incidentRole } = req.params
+    const { selectedDecisionId } = req.body
+
+    const error = validateForm({ selectedDecisionId })
```
richardpopple, line 64: Still plenty of stuff to do regarding getting additional info e.g. questions where we search for other prisoners

richardpopple: I figure that we'll stick that in the session - its possible we could do it with url parameters but I'm thinking that way will be tricky....

server/views/pages/offenceCodeDecisions.njk
```diff
@@ -0,0 +1,46 @@
...
+            {
+                value: question.id,
+                text: question.label
+            }
+        ), items) %}
+    {% endfor %}
+
+    <form method="POST" novalidate="novalidate" class="govuk-!-margin-top-5">
+        <input type="hidden" name="_csrf" value="{{ csrfToken }}" />
+        {{ govukRadios({
```
richardpopple, line 28: Again still lots to do with the special questions.

package.json
```diff
@@ -135,17 +135,17 @@
     "@types/redis": "^2.8.32",
     "@types/superagent": "^4.1.13",
     "@types/supertest": "^2.0.11",
-    "@typescript-eslint/eslint-plugin": "^4.33.0",
-    "@typescript-eslint/parser": "^4.33.0",
+    "@typescript-eslint/eslint-plugin": "^5.10.2",
```
richardpopple, line 138: Upped the version of the linter in the hope that it would fix emum issues - it didn't but I figure being at the latest is probably worth leaving in

server/offenceCodeDecisions/Commited.ts
```diff
@@ -0,0 +1,156 @@
+import Decision from './Decision'
```
sp-watson, line 1: typo in class name

server/offenceCodeDecisions/Commited.ts
```diff
@@ -0,0 +1,156 @@
...
+// Codes
+const todo = new Code('TODO')
+
+function decision(question: Question | string) {
+  return new Decision(question)
+}
+
+export default new Decision()
+  .title(
+    `What type of offence did ${PlaceholderText.OFFENDER_FIRST_NAME} ${PlaceholderText.OFFENDER_FIRST_NAME} commit?`
```
sp-watson, line 15: Hmm.. I would say it's one item - PlaceholderText.OFFENDER_FULL_NAME? Anyway, it\'s wrong as 2 first names

server/offenceCodeDecisions/Commited.ts
sp-watson: is there a reason why you don't use PlaceholderText (or in fact, is PlaceholderText required at all)?

server/routes/offenceCodeDecisions/index.ts
sp-watson: i think it would be better to get all decisions and construct routes for each in a loop

server/offenceCodeDecisions/Commited.ts
richardpopple: So this was tied up with not being able to get enums working - however I think I've fixed that now and we use the enums in the titles here.

server/routes/offenceCodeDecisions/index.ts
richardpopple: So instead of using the star matching?

server/offenceCodeDecisions/Commited.ts
richardpopple: fixed

server/offenceCodeDecisions/Commited.ts
richardpopple: I did think about that, but I though that the rest of the application has a mixture of LAST, FIRST and FIRST, LAST and was thinking we might see this here. Can change if you want though?

richardpopple: Actually - I've made the change, but kept the functionality to allow us to be more specific if we need to be I think

server/routes/offenceCodeDecisions/index.ts
richardpopple: So I've done this and I think it looks quite a bit better! Thank you!

server/utils/nunjucksSetup.ts
```diff
@@ -50,7 +50,7 @@ export default function nunjucksSetup(app: express.Express, path: pathModule.Pla
     return `${array[0][0]}. ${array.reverse()[0]}`
   })
 
-  njkEnv.addFilter('findError', (array: FormError[] = [], formFieldId: string) => {
```
richardpopple, line 53: linter doesn't like this

server/services/healthCheck.test.ts
```diff
@@ -36,22 +36,22 @@ describe('Healthcheck', () => {
 
 function successfulCheck(name: string): HealthCheckService {
   return () =>
-    new Promise((resolve, _reject) =>
```
richardpopple, line 39: linter didn't like this

server/offenceCodeDecisions/Decision.ts
sp-watson: Yep - sounds ok. We only care about the end result, which is the `code`.

server/offenceCodeDecisions/Decision.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  getPage(): string {
+    return this.decisionPage
+  }
+
+  getCode() {
+    return this.decisionCode
+  }
+
+  allUrls(): Array<string> {
```
sp-watson, line 101: What's this?! Is it the path to the given question?

server/offenceCodeDecisions/Decision.ts
```diff
@@ -0,0 +1,191 @@
...
+    const codes = [].concat(...this.getChildren().map(c => c.allUrls()))
+    if (this.getUrl()) {
+      codes.push(this.getUrl())
+    }
+    return codes.sort()
+  }
+
+  findByUrl(url: string): Decision {
+    return this.findBy(d => d.getUrl() === url)
```
sp-watson, line 110: Do we need this.. I think so, just a note to check

server/offenceCodeDecisions/Decision.ts
sp-watson: Hmm.. I would have envisaged that it would just require a code to get the questions.

server/offenceCodeDecisions/Decision.ts
sp-watson: yeah, good.. but where would this be used?

server/offenceCodeDecisions/DecisionTrees.test.ts
```diff
@@ -0,0 +1,44 @@
+import decisionTrees from './DecisionTrees'
```
sp-watson, line 1: I don't see a test to make sure codes are only assigned to the leafs?

server/offenceCodeDecisions/DecisionTrees.test.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+describe('decisions', () => {
+  it.skip('toString', () => {
+    // Not a test but useful output
+    // eslint-disable-next-line no-console
+    decisionTrees.forEach(decisionTree => console.log(decisionTree.toString()))
+  })
+
+  it.skip('no duplicate codes', () => {
```
sp-watson, line 23: Why skipped? is it because codes can be duplicated, because i think that is correct.....

we have situations where an answer before the end doesn't actually change the output code (so the same branch), but that is information that is sent to the API.

server/offenceCodeDecisions/Placeholder.ts
```diff
@@ -0,0 +1,21 @@
...
+  offenderFirstName: string
+  offenderLastName: string
+}
+
+// eslint-disable-next-line no-shadow
+export const enum PlaceholderText {
+  OFFENDER_FULL_NAME = '{OFFENDER_FULL_NAME}',
+  OFFENDER_FIRST_NAME = '{OFFENDER_FIRST_NAME}',
```
sp-watson, line 9: Are these actually used?

server/offenceCodeDecisions/Placeholder.ts
```diff
@@ -0,0 +1,21 @@
...
+  OFFENDER_LAST_NAME = '{OFFENDER_LAST_NAME}',
+}
+
+export function getProcessedText(template: string, placeholderValues: PlaceholderValues): string {
+  return (template || '')
+    .replace(
+      PlaceholderText.OFFENDER_FULL_NAME,
+      `${placeholderValues.offenderLastName} ${placeholderValues.offenderLastName}`
```
sp-watson, line 17: last name twice/

server/routes/index.ts
sp-watson: Look Ok to me! i suppose `offence-code-selection` might be a bit more appropriate?!

server/routes/offenceCodeDecisions/index.ts
```diff
@@ -0,0 +1,23 @@
...
+export default function offenceCodeDecisionsRoutes({
+  placeOnReportService,
+}: {
+  placeOnReportService: PlaceOnReportService
+}): Router {
+  const router = express.Router()
+  const offenceCodeDecisions = new OffenceCodeDecisionsRoutes(placeOnReportService)
+  const get = (path: string, handler: RequestHandler) => router.get(path, asyncMiddleware(handler))
+  const post = (path: string, handler: RequestHandler) => router.post(path, asyncMiddleware(handler))
+  decisionTrees.forEach((decisionTree, incidentRole) =>
```
sp-watson, line 16: Beautiful.. except the use of url. i would maybe go through the decision object themselves

server/routes/offenceCodeDecisions/offenceCodeDecisions.ts
```diff
@@ -0,0 +1,83 @@
...
+    return {
+      offenderFirstName: properCaseName(prisonerDetails.firstName),
+      offenderLastName: properCaseName(prisonerDetails.lastName),
+    }
+  }
+
+  private renderView = async (req: Request, res: Response, pageData?: PageData): Promise<void> => {
+    const { adjudicationNumber, incidentRole } = req.params
+    const { error, selectedDecisionId } = pageData
+    const placeholderValues = await this.placeholderValues(adjudicationNumber, res)
```
sp-watson, line 37: We ought to only get this info when we need it.. but not easy at all!

server/routes/offenceCodeDecisions/offenceCodeDecisions.ts
sp-watson: Yeah session.. the only other options are POSTing and cookies, but let's discuss that later

server/routes/offenceCodeDecisions/offenceCodeDecisions.ts
sp-watson: Yep

server/routes/offenceCodeDecisions/offenceCodeDecisionsValidation.ts
sp-watson: LGTM

server/offenceCodeDecisions/Decision.ts
sp-watson: Good for logging - log at info level on entry to page

server/offenceCodeDecisions/Decision.ts
sp-watson: Yes, used for routing, express does parsing

server/offenceCodeDecisions/DecisionTrees.test.ts
sp-watson: Covered by invalid decision code

server/offenceCodeDecisions/DecisionTrees.test.ts
sp-watson: Until we do todos..

server/routes/offenceCodeDecisions/index.ts
sp-watson: It's good - using express parser

sp-watson (approved):

PR #128, opened 2022-02-16
NN-3805: Delete person page and tests

natclamp-moj:


integration_tests/integration/incidentDetailsEdit.spec.ts
```diff
@@ -318,13 +318,12 @@ context('Incident details (edit) - statement incomplete', () => {
     incidentDetailsPage.timeInputMinutes().type('00')
     incidentDetailsPage.locationSelector().select('Workshop 2')
     incidentDetailsPage.inciteAssociatedPrisonerDeleteButton().click()
-    // TODO: Once the /delete-person page is created, it will redirect automatically to this page so the next line can be removed
-    cy.visit(`/incident-details/G6415GD/34/edit`)
+    cy.get('[data-qa="radio-buttons"]').find('input[value="yes"]').check()
+    cy.get('[data-qa="delete-person-submit"]').click()
     incidentDetailsPage.timeInputHours().should('have.value', '13')
```
sp-watson, line 323: This seems to be missing the check of the loc.pathname and loc.search, as is done in `incidentDetailSubmiitedEdit.spec.ts`

server/routes/incidentDetails/incidentDetailsEdit.ts
```diff
@@ -46,10 +46,12 @@ export default class IncidentDetailsEditRoutes {
   private getAssociatedPrisonerNumber = (
     selectedRadio: string,
     previouslyReportedAssociatedPrisoner: string,
-    newAssociatedPrisoner: string
+    newAssociatedPrisoner: string,
+    personDeleted: string
   ): string => {
     if (selectedRadio === 'onTheirOwn' || selectedRadio === 'attemptOnTheirOwn') return null
-    if (newAssociatedPrisoner !== undefined) return newAssociatedPrisoner
+    if (personDeleted === 'true') return null
```
sp-watson, line 53: Do we actually need this (prob do, but can't quite see why).. if the selectedPerson is always set then we just need to check whether that exists, don't we?

server/routes/incidentDetails/incidentDetailsEdit.ts
```diff
@@ -135,17 +142,26 @@ export default class IncidentDetailsEditRoutes {
     } = req.body
     const { user } = res.locals
     const { prisonerNumber, id } = req.params
+    const { personDeleted } = req.query
```
sp-watson, line 145: Associated with other comment - do we need this?

server/routes/incidentDetails/incidentDetailsEdit.ts
```diff
@@ -135,17 +142,26 @@ export default class IncidentDetailsEditRoutes {
...
     const selectedPerson = JSON.stringify(req.query.selectedPerson)?.replace(/"/g, '')
     const { originalRadioSelection } = req.session
 
+    const newlySelectedAssociatedPrisonersNumber =
```
sp-watson, line 149: Minor thing, and it might not be possible, but we should probably have the 1search` code above this as it does not need this info, does it?

server/routes/incidentDetails/incidentDetailsEdit.ts
```diff
@@ -187,7 +196,6 @@ export default class IncidentDetailsEditRoutes {
         incidentDate,
         locationId,
         originalRadioSelection: currentRadioSelected,
-        associatedPrisonersNumber,
```
sp-watson, line 190: A bit confused as to why this is no longer required?!

sp-watson (approved):

PR #131, opened 2022-02-21
NN-3812: Update task list with offence details

natclamp-moj:


server/services/placeOnReportService.ts
```diff
@@ -247,6 +247,7 @@ export default class PlaceOnReportService {
       handoverDeadline: draftAdjudication.incidentDetails.handoverDeadline,
       statementPresent: !!draftAdjudication.incidentStatement,
       statementComplete: draftAdjudication.incidentStatement?.completed || false,
+      offenceDetailsComplete: draftAdjudication.offenceDetails?.length > 0,
```
natclamp-moj, line 250: I'm under the impression that the offence details array will either be not present at all, or have at least 1 element (on the draft adjudication)

server/data/DraftAdjudicationResult.ts
```diff
@@ -14,6 +14,19 @@ export type IncidentStatement = {
   completed?: boolean
 }
 
+export type OffenceRule = {
```
sp-watson, line 17: I think you are going to have a conflict with Rich's stuff - he has added (some) of this

server/services/placeOnReportService.ts
sp-watson: Yep - exactly - and I assume having 1 or more means it is complete?!

sp-watson (approved):

PR #147, opened 2022-02-25
NN-3935: offence question data structure part 4

richardpopple:


server/routes/detailsOfOffence/detailsOfOffence.ts
```diff
@@ -74,6 +74,13 @@ export default class DetailsOfOffenceRoutes {
     return res.redirect(`/incident-statement/${prisonerNumber}/${adjudicationNumber}`)
   }
 
+  deleteOffence = async (req: Request, res: Response): Promise<void> => {
+    const index = Number(req.params.index)
+    const adjudicationNumber = Number(req.params.adjudicationNumber)
+    this.allOffencesSessionService.deleteSessionOffence(req, index, adjudicationNumber)
```
sp-watson, line 80: I expected to see a `await this.populateSessionIfEmpty(adjudicationNumber, req, res)` as per the `addOffence` but maybe it's not necessary?

richardpopple: So this is just deleting from the session - it doesn't call the api so it doesn't need an await

sp-watson: ah right - I misunderstood - that method is getting the draft data and adding to the session. makes sense now

sp-watson (approved):

PR #152, opened 2022-03-02
NN-3935: offence question data structure part 6

richardpopple:


server/offenceCodeDecisions/Decision.ts
```diff
@@ -86,15 +86,16 @@ export class Decision {
     return this.findDecisionBy(d => d.getUrl() === url)
   }
 
+  findDecisionById(id: string): Decision {
```
sp-watson, line 89: This is function just for tests - consider moving

sp-watson (approved):

server/offenceCodeDecisions/Decision.ts
richardpopple: Yeah - I'll pull that out

PR #165, opened 2022-03-17
NN-3950: Incident details pages refactor

sp-watson:


integration_tests/integration/incidentDetailsEdit.spec.ts
```diff
@@ -262,7 +262,8 @@ context('Incident details (edit) - statement incomplete', () => {
       expect(loc.pathname).to.eq('/offence-code-selection/34/assisted/1')
     })
   })
-  it('should submit form successfully if all data entered and redirect to /offence-details page - change time', () => {
+  // TODO - To be changed - goes to task list
```
sp-watson, line 265: These TODOs are what we will address after tomorrow's meeting

server/routes/incidentDetails/incidentDetails.ts
```diff
@@ -1,190 +1,20 @@
 import { Request, Response } from 'express'
```
sp-watson, line 1: The 3 original pages are exactly the same now, except for the PageRequestType

.eslintrc.json
```diff
@@ -33,6 +33,10 @@
         "plugin:prettier/recommended"
       ],
       "rules": {
+        // fix for enum failures https://github.com/typescript-eslint/typescript-eslint/issues/2484
```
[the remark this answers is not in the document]
richardpopple: If that's cool then I can remove a bunch of inline annotations for enums in my code!

server/routes/incidentDetails/incidentDetails.ts
richardpopple: Not sure if it makes any difference - but could you have a single class and then get the route to construct it with the enum - that way you wouldn't need the three files?

server/routes/incidentDetails/incidentDetailsPage.ts
```diff
@@ -0,0 +1,728 @@
...
+    private readonly locationService: LocationService
+  ) {
+    this.pageOptions = new PageOptions(pageType)
+  }
+
+  view = async (req: Request, res: Response): Promise<void> => {
+    const requestValues = extractValuesFromRequest(req)
+    const requestData = getPageActionFromRequest(requestValues)
+    const { user } = res.locals
+    debugData('GET', requestValues)
```
richardpopple, line 136: These staying in?

server/routes/incidentDetails/incidentDetailsPage.ts
```diff
@@ -0,0 +1,728 @@
...
+      }
+    }
+    debugData('GET DATA', data)
+
+    let originalReporterUsernameForPage = user.username
+    if (this.pageOptions.isEdit()) {
+      originalReporterUsernameForPage = data.temporaryData?.originalReporterUsername
+    }
+    const pageData = await this.getPageDataOnGet(requestValues, originalReporterUsernameForPage, data, user)
+    renderData(res, pageData, null)
```
richardpopple, line 164: Normally we return renderData as a call? Maybe it doesn't matter? Just wondering if the deleteSessionData should be called before this and then we return it?

server/routes/incidentDetails/incidentDetailsPage.ts
```diff
@@ -0,0 +1,728 @@
...
+
+  view = async (req: Request, res: Response): Promise<void> => {
+    const requestValues = extractValuesFromRequest(req)
+    const requestData = getPageActionFromRequest(requestValues)
+    const { user } = res.locals
+    debugData('GET', requestValues)
+    debugData('GET action', requestData)
+
+    let data: StashedIncidentDetails = null
+    if (requestData !== PageRequestAction.STANDARD) {
```
richardpopple, line 140: Cool - like that its clear whats going on here

server/routes/incidentDetails/incidentDetailsPage.ts
```diff
@@ -0,0 +1,728 @@
...
+    })
+    if (validationError) {
+      // Could stash to session and redirect here
+      const pageData = await this.getPageDataOnPost(postValues, user)
+      return renderData(res, pageData, validationError)
+    }
+
+    const incidentDetailsToSave = postValues.incidentDetails
+
+    try {
```
richardpopple, line 224: Out of interest - why would we get an exception here? Should we just let it throw and get the sorry screen?

server/routes/incidentDetails/incidentDetailsPage.ts
```diff
@@ -0,0 +1,728 @@
...
+    return PageRequestAction.SEARCH_FOR_PRISONER
+  }
+  if (reqData.deleteWanted) {
+    return PageRequestAction.DELETE_PRISONER
+  }
+  return PageRequestAction.STANDARD
+}
+
+const popDataFromSession = (req: Request): StashedIncidentDetails => {
+  const { incidentDate } = req.session
```
richardpopple, line 414: Maybe for another time - but is it worth having an object on the session instead of distinct properties?

server/routes/incidentDetails/incidentDetailsPage.ts
```diff
@@ -0,0 +1,728 @@
...
+) => {
+  return res.redirect(`/check-your-answers/${prisonerNumber}/${draftId}/${isReviewerPage ? 'review' : 'report'}`)
+}
+
+const redirectToTaskList = (res: Response, prisonerNumber: string, draftId: number) => {
+  return res.redirect(getTaskListUrl(prisonerNumber, draftId))
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const setUpRedirectForEditError = (res: Response, prisonerNumber: string, error: any, draftId: number) => {
```
richardpopple, line 701: Could error be typed as an Error? or is potentially something else?

server/routes/incidentDetails/incidentDetailsPage.ts
```diff
@@ -0,0 +1,728 @@
...
+    case IncidentRole.ASSISTED:
+    // Fall through
+    default:
+      radioSelectionCode = 'assisted'
+      break
+  }
+  return radioSelectionCode
+}
+
+const incidentRoleFromRadioSelection = (radioSelectionCode: string): IncidentRole => {
```
richardpopple, line 640: I think this *might* be possible without the if statement - the IncidentRole enum has these values encoded in them - it feels like it should be able to reverse. I might have a play

richardpopple (approved):

.eslintrc.json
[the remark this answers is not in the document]
sp-watson: Yeah I think so.. i don't really get the no-shadow check anyway... I guess it makes it harder to understand if there is >1 symbol with the same name?

server/routes/incidentDetails/incidentDetails.ts
sp-watson: That is a decent idea - I might do that later

server/routes/incidentDetails/incidentDetailsPage.ts
sp-watson: I was preserving the existing functionality - but it does things like set up a specific URL for the Sorry page on error

server/routes/incidentDetails/incidentDetailsPage.ts
sp-watson: Ha yeah - that does make sense!

server/routes/incidentDetails/incidentDetailsPage.ts
sp-watson: I couldn't see anything specific in the catch documentation. Suppose I could debug it?!

PR #167, opened 2022-03-18
NN-3976: Rationalise Check your answers pages

natclamp-moj:


server/routes/checkYourAnswers/checkYourAnswersPage.ts
```diff
@@ -0,0 +1,153 @@
...
+    private readonly decisionTreeService: DecisionTreeService
+  ) {
+    this.pageOptions = new PageOptions(pageType)
+  }
+
+  private renderView = async (req: Request, res: Response, pageData: PageData): Promise<void> => {
+    const { error } = pageData
+    const { user } = res.locals
+
+    const { adjudicationNumber, draftAdjudication, incidentRole, prisoner, associatedPrisoner } =
```
sp-watson, line 91: Could put these in a Promise.all

server/routes/checkYourAnswers/checkYourAnswersPage.ts
```diff
@@ -0,0 +1,153 @@
...
+    pageType: PageRequestType,
+    private readonly placeOnReportService: PlaceOnReportService,
+    private readonly locationService: LocationService,
+    private readonly decisionTreeService: DecisionTreeService
+  ) {
+    this.pageOptions = new PageOptions(pageType)
+  }
+
+  private renderView = async (req: Request, res: Response, pageData: PageData): Promise<void> => {
+    const { error } = pageData
```
sp-watson, line 88: Just a suggestion, but I converted the req and res stuff to data structures at the beginning (and the end) of the main methods (view and submit). That way it makes it clear what is expected from the req/res and what we are putting on it.

Having said all that, this is pretty clear in this page, so it's not that important.

PR #176, opened 2022-03-23
NN-3992: Remove PRNs from statement and check answers

natclamp-moj:


server/routes/incidentDetails/incidentDetailsPage.ts
```diff
@@ -707,6 +707,7 @@ const setUpRedirectForEditError = (res: Response, prisonerNumber: string, error:
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const setUpRedirectForCreationError = (res: Response, prisonerNumber: string, error: any) => {
   logger.error(`Failed to post incident details for draft adjudication: ${error}`)
+  // TODO: @steve not sure this URL is correct - shouldn't it be /id ?
   res.locals.redirectUrl = `/incident-statement/${prisonerNumber}`
```
sp-watson, line 711: Yep - I copied this stuff from the original so i think it was always wrong.

sp-watson (approved):

server/routes/incidentDetails/incidentDetailsPage.ts
natclamp-moj: I'll change it in the next PR - obviously wasn't used very much if it was always wrong and nobody noticed!

PR #186, opened 2022-03-28
NN-3816 printed form view part 2

richardpopple:


server/routes/printReport/printReport.ts
```diff
@@ -37,7 +41,21 @@ export default class PrintReportRoutes {
     const { user } = res.locals
     const { pdfMargins, adjudicationsUrl } = config.apis.gotenberg
     const adjudicationDetails = await this.reportedAdjudicationsService.getConfirmationDetails(adjudicationNumber, user)
-    const noticeOfBeingPlacedOnReportData = new NoticeOfBeingPlacedOnReportData(adjudicationNumber, adjudicationDetails)
+
+    const { reportedAdjudication, associatedPrisoner, prisoner } =
+      await this.decisionTreeService.reportedAdjudicationIncidentData(adjudicationNumber, user)
+    const offences = await this.decisionTreeService.getAdjudicationOffences(
```
sp-watson, line 47: A bit confused by this - the offences are available from the API call. is this for the questions and answers?

server/services/decisionTreeService.ts
```diff
@@ -2,99 +2,115 @@ import { Decision } from '../offenceCodeDecisions/Decision'
...
+    return {
+      draftAdjudication,
+      incidentRole: data.incidentRole,
+      associatedPrisoner: data.associatedPrisoner,
+      prisoner: data.prisoner,
+    }
+  }
+
+  async reportedAdjudicationIncidentData(adjudicationNumber: number, user: User) {
+    const { reportedAdjudication } = await this.reportedAdjudicationService.getReportedAdjudicationDetails(
```
sp-watson, line 48: Could do with confirmation of how this is working.

server/services/reportedAdjudicationsService.ts
```diff
@@ -25,6 +29,10 @@ export default class ReportedAdjudicationsService {
     private readonly locationService: LocationService
   ) {}
 
+  async getReportedAdjudicationDetails(adjudicationNumber: number, user: User): Promise<ReportedAdjudicationResult> {
+    return new ManageAdjudicationsClient(user.token).getReportedAdjudication(adjudicationNumber)
```
sp-watson, line 33: What's the difference between this and the below function?

server/routes/printReport/printReport.ts
richardpopple: Yeah - all of this is to get the questions and answers (and fill them out with data - like prisoner names and whatnot)

server/services/reportedAdjudicationsService.ts
richardpopple: so the getConfirmationDetails pulls a bunch of stuff together - like neurodiversity etc. which we need for the non print page. And the getReportedAdjudication just brings back the adjudication - which we need for the actual printed page. I think we might be able to get away with not calling the getConfirmationDetails for the rendering of the pdf - but that will come in a later PR I think

sp-watson (approved):

PR #196, opened 2022-04-07
NN-4045: Tests to use url generator

natclamp-moj:


server/data/ReportedAdjudicationResult.ts
```diff
@@ -22,5 +22,4 @@ export interface ReportedAdjudicationEnhanced extends ReportedAdjudication {
   formattedDateTimeOfIncident: string
   dateTimeOfIncident: string
   reportingOfficer?: string
-  prisonerReportUrl?: string
```
natclamp-moj, line 25: removed this like we spoke about @richardpopple

integration_tests/integration/checkYourAnswers.spec.ts
```diff
@@ -137,7 +144,7 @@ context('Check Your Answers', () => {
     cy.signIn()
   })
   it('should contain the required page elements', () => {
-    cy.visit(`/check-your-answers/3456`)
+    cy.visit(`${checkYourAnswers.urls.start(3456)}`)
```
richardpopple, line 147: do you need to have these in quotes?

server/data/ReportedAdjudicationResult.ts
richardpopple: This now a parameter passed in from the controller?

richardpopple (approved):

PR #198, opened 2022-04-08
NN-4052: Nunjucks filters for url generation

natclamp-moj:


server/utils/nunjucksSetup.ts
```diff
@@ -121,5 +132,18 @@ export default function nunjucksSetup(app: express.Express, path: pathModule.Pla
   njkEnv.addGlobal('authUrl', config.apis.hmppsAuth.url)
   njkEnv.addGlobal('digitalPrisonServiceUrl', config.digitalPrisonServiceUrl)
   njkEnv.addGlobal('supportUrl', config.supportUrl)
+  njkEnv.addGlobal('homepageUrl', homepageUrl.root)
```
richardpopple, line 135: Not sure if its possible - but would a single import here work - so if we scoped them all to be inside a single object
const our_urls = {
   offenceCodeSelection:  {....} 
   deleteOffence: {...}
}
And then have:
njkEnv.addFilter('our_urls', our_urls)
(Not sure if this would work - as you'd be returning an object and not a function)
But if it did then it would mean we wouldn't need to add to this file when adding new urls.  
Might be worth some investigation? Happy to talk about it

server/utils/urlGenerator.ts
```diff
@@ -183,3 +183,10 @@ export const deletePerson = {
     start: '/',
   },
 }
+
```
richardpopple, line 186: This should probably be called homepage so as to be consistent with the others in this file?

richardpopple (approved):

PR #208, opened 2022-04-13
NN-4009 read only view

timoj58:


server/routes/prisonerReport/prisonerReportPage.ts
```diff
@@ -110,6 +112,7 @@ export default class prisonerReportRoutes {
       offences,
       statementEditable: !this.pageOptions.isReviewerView(),
       ...prisonerReportVariables,
+      readOnly: ['ACCEPTED', 'REJECTED'].includes(reportedAdjudication.reportedAdjudication.status),
```
richardpopple, line 115: Maybe we stick these in an enum - that mirrors the enum on the API. I might have a look in the stuff I'm doing

richardpopple (approved):

PR #210, opened 2022-04-20
NN-4001 your completed reports page updates

richardpopple:


assets/sass/_horizontal-content.scss
```diff
@@ -1,45 +1,3 @@
-.horizontal-information {
```
richardpopple, line 1: None of this was referenced - was pulled in from restricted patients

assets/sass/_horizontal-content.scss
```diff
@@ -57,23 +15,26 @@
...
 
     button,
     a {
       align-self: flex-end;
-      margin-bottom: $govuk-border-width-form-element;
     }
   }
 }
+
+.horizontal-form-container {
```
richardpopple, line 35: So in this instance we want a grey background

server/utils/nunjucksSetup.ts
```diff
@@ -77,16 +77,16 @@ export default function nunjucksSetup(app: express.Express, path: pathModule.Pla
     return null
   })
 
-  njkEnv.addFilter('toSelect', (array, valueKey, textKey, value) => {
+  njkEnv.addFilter('toSelect', (array, valueKey, textKey, value, emptyOptionText) => {
```
richardpopple, line 80: We now allow string values - not just numbers or string representation of numbers

integration_tests/mockApis/adjudications.ts
```diff
@@ -179,21 +181,31 @@ const stubGetYourReportedAdjudications = ({
   number = 0,
   size = 20,
   allContent = [],
+  filter = { status: null, toDate: moment().format('YYYY-MM-DD'), fromDate: moment().format('YYYY-MM-DD') },
```
sp-watson, line 184: Using the current date is a bit risky for tests - can they be set to null or in fact any old value?

Actually, maybe just don't have a default at all?!

server/utils/urlGenerator.ts
```diff
@@ -118,6 +118,9 @@ const adjudicationUrls = {
   },
   yourCompletedReports: {
     root: '/your-completed-reports',
+    urls: {
```
sp-watson, line 121: is this needed? I can't see where it is used

server/routes/allCompletedReports/allCompletedReports.test.ts
```diff
@@ -42,7 +43,8 @@ beforeEach(() => {
       },
       incidentRole: {},
       offenceDetails: [{ offenceCode: 1001 }],
-      status: 'AWAITING_REVIEW',
+      status: ReportedAdjudicationStatus.AWAITING_REVIEW,
+      statusDisplayName: 'AWAITING_REVIEW',
```
sp-watson, line 47: Not sure this is right - I thought it should be e.g. "Awaiting review"?

server/services/reportedAdjudicationsService.test.ts
```diff
@@ -138,7 +141,7 @@ describe('reportedAdjudicationsService', () => {
     })
 
     it('returns the correct data', async () => {
-      const result = await service.getYourCompletedAdjudications(user, {
+      const result = await service.getYourCompletedAdjudications(user, null, {
```
sp-watson, line 144: I think we need a test for the filters as well (but that can be added in another PR)

server/routes/yourCompletedReports/yourCompletedReports.ts
```diff
@@ -1,30 +1,72 @@
...
+  submit = async (req: Request, res: Response): Promise<void> => {
+    const urlQuery = {
+      pathname: adjudicationUrls.yourCompletedReports.root,
+      query: {
+        fromDate: req.body.fromDate.date,
+        toDate: req.body.toDate.date,
+        status: req.body.status as ReportedAdjudicationStatus,
+      },
+    }
+    return res.redirect(url.format(urlQuery))
```
sp-watson, line 70: yeah good - I guess a redirect is the only way as they need to re-submit

richardpopple: This is also the only way I think it can work with the pagination.....

server/routes/allCompletedReports/allCompletedReports.test.ts
richardpopple: You are correct - I'll change

integration_tests/mockApis/adjudications.ts
richardpopple: So... the reason for this is that today is the default - so when we call the the api we would expect to get these values requested.

server/utils/urlGenerator.ts
richardpopple: I think you are correct - I'll remove

server/services/reportedAdjudicationsService.test.ts
richardpopple: Yeah you are correct - I figure I'll put this in when doing the report adjudications stuff.

server/utils/urlGenerator.ts
richardpopple: Ah - actually its used in the njk file

sp-watson (approved):

sp-watson (approved):

sp-watson (approved):

PR #223, opened 2022-04-27
NN-4002 bug fixes

timoj58:


server/views/partials/reviewPanelRadios.njk
```diff
@@ -26,9 +25,12 @@
                value: "expired",
                text: "More than 48 hours have elapsed since the incident"
              }
-          ] | toSelect('value', 'text'),
+          ] | toSelect('value', 'text', pageData.reason),
 					classes: '',
-    			errorMessage: 'rejectedReasonId' | findError(pageData.errors)
+          attributes: {
```
sp-watson, line 30: This formatting is not good - we really have to review this formatter you are using.

For one, why is it using tabs? Also, is this a good way of formatting - to have newlines for single attributes?

integration_tests/integration/prisonerReportReview.spec.ts
```diff
@@ -321,4 +359,19 @@ context('Prisoner report - reviewer view', () => {
     PrisonerReportPage.reviewSubmit().click()
     cy.get('*[class^="govuk-error-message"]').contains('A reason is required')
   })
+  it('should not contain the review panel if status is RETURNED', () => {
```
sp-watson, line 362: Let's just hold this off for a minute... we need to check this with the designers as well

server/routes/prisonerReport/prisonerReportReviewValidation.ts
```diff
@@ -43,18 +43,23 @@ export const errors: { [key: string]: FormError } = {
...
+
+    if (returnedErrors.length !== 0) return returnedErrors
+
+    if (details.length > 4000) return [errors.REJECT_DETAILS_WORD_COUNT_EXCEEDED]
   }
   if (status === ReviewStatus.RETURNED) {
-    if (!reason) return errors.MISSING_RETURN_REASON
-    if (!details) return errors.MISSING_RETURN_DETAILS
-    if (details.length > 4000) return errors.RETURN_DETAILS_WORD_COUNT_EXCEEDED
+    if (!reason) returnedErrors.push(errors.MISSING_RETURN_REASON)
```
sp-watson, line 59: Just noticed these 4 lines are virtually the same as above. 

Personally I don't like this much duplication - but it isn't a big issue

server/routes/prisonerReport/prisonerReportReviewValidation.ts
```diff
@@ -43,18 +43,23 @@ export const errors: { [key: string]: FormError } = {
...
 
+  const returnedErrors: FormError[] = []
   if (status === ReviewStatus.REJECTED) {
-    if (!reason) return errors.MISSING_REJECT_REASON
-    if (!details) return errors.MISSING_REJECT_DETAILS
-    if (details.length > 4000) return errors.REJECT_DETAILS_WORD_COUNT_EXCEEDED
+    if (!reason) returnedErrors.push(errors.MISSING_REJECT_REASON)
+    if (!details) returnedErrors.push(errors.MISSING_REJECT_DETAILS)
+
+    if (returnedErrors.length !== 0) return returnedErrors
```
sp-watson, line 54: I would prefer > 0 - just a bit clearer for me. But it's your call

integration_tests/integration/prisonerReportReview.spec.ts
sp-watson: Approved!

server/views/partials/reviewPanelRadios.njk
timoj58: that formatting is github, its fine in the code

sp-watson (approved):

PR #227, opened 2022-04-28
NN-4080 validate dates

richardpopple:


server/routes/yourCompletedReports/yourCompletedReports.ts
```diff
@@ -1,56 +1,57 @@
...
 
   submit = async (req: Request, res: Response): Promise<void> => {
-    return res.redirect(
-      adjudicationUrls.yourCompletedReports.urls.filter(
-        req.body.fromDate.date,
-        req.body.toDate.date,
-        req.body.status as ReportedAdjudicationStatus
-      )
-    )
+    const uiFilter = uiFilterFromBody(req)
```
sp-watson, line 50: I like this explicit conversion to a type - i think it i much more readable actually!

sp-watson (approved):

PR #233, opened 2022-05-03
NN-4090 add card script

timoj58:


sp-watson (approved):

server/views/partials/layout.njk
```diff
@@ -58,4 +58,8 @@
   <script src="/assets/govukFrontendInit.js"></script>
   <script src="/assets/moj/all.js"></script>
   <script src="/assets/mojFrontendInit.js"></script>
+  {% block pageScripts %}
+     <script src="/assets/card.js"></script>
```
sp-watson, line 62: This explicit reference is not needed - the `block` directive includes the code in the block in the njk template that uses it.

So you just need:

```
{% block pageScripts %}
{%  endblock %}
```

natclamp-moj, line 62: I think this can just be an empty block i.e. 
{% block pageScripts %}
{%  endblock %}

as we load the card.js file into the block in the homepage.njk don't we? We don't necessarily want the card.js file being loaded on every page which is what this is doing

sp-watson (approved):

natclamp-moj (approved):

PR #240, opened 2022-05-09
NN-4062

timoj58:
refactor test as per ticket


integration_tests/integration/incidentDetails.spec.ts
```diff
@@ -1,17 +1,11 @@
...
 import adjudicationUrls from '../../server/utils/urlGenerator'
 
-// In order to bypass the date picker we force the input to accept text and then press escape so the date picker
-// disappears allowing us to interact with other fields.
-const forceDateInput = (day: number, month: number, year: number): PageElement =>
-  cy
-    .get('[data-qa="incident-details-date"]')
-    .clear({ force: true })
-    .type(`${`0${day}`.slice(-2)}/${`0${month}`.slice(-2)}/${year}{esc}`, { force: true })
+const month = new Date().getUTCMonth() + 1
```
sp-watson, line 8: Probably not worth revisiting, but this month+1 should probably be done in the `forceDateInput`, so the date would be passed in to that function instead of the day, month and year.

It's up to you whether to change it

integration_tests/componentDrivers/DateInput.ts
```diff
@@ -0,0 +1,18 @@
+import { PageElement } from '../pages/page'
```
sp-watson, line 1: A naming convention thing: All our files start with a lower case.

integration_tests/integration/incidentDetails.spec.ts
timoj58: the reason i left it outside, is that people also pass actual string dates, ie "10" therefore we would be subtracting one of that.  Which is wrong.    One option would be to create a new Date only function.

sp-watson (approved):

PR #247, opened 2022-05-20
NN-4111: Add gtag

natclamp-moj:


assets/js/reviewerAction.js
```diff
@@ -0,0 +1,30 @@
+$(function () {
+  const form = document.querySelector('form')
+  form.addEventListener('submit', e => {
+    e.preventDefault()
```
sp-watson, line 4: Doesn't this stop the submit from working?

Could we have an empty callback and not have to `preventDefault`?

assets/js/reviewerAction.js
```diff
@@ -0,0 +1,30 @@
+$(function () {
```
sp-watson, line 1: The filename needs to be more descriptive, i.e. it needs to show that it is for reporting purposes only. How about `reviewerAnalytics`?

assets/js/reviewerAction.js
```diff
@@ -0,0 +1,30 @@
...
+  const form = document.querySelector('form')
+  form.addEventListener('submit', e => {
+    e.preventDefault()
+    const radio = document.querySelector('[class="govuk-radios__item"] :checked')
```
sp-watson, line 5: This is good logic, the trouble is it relies entirely on id's elsewhere with no connection between the two of them. 

Worse still, this stuff will not give us any errors if it stops working unless we do something else, like write an int test.

Can we write an int test for this? If not, we might have to look at adding something to the submit to make it obvious to the person dealing with that page that this will break if they change it

PR #250, opened 2022-06-01
NN-4110: Update node and npm

natclamp-moj:


package.json
```diff
@@ -28,8 +28,8 @@
     "clean": "rm -rf dist build node_modules stylesheets"
   },
   "engines": {
-    "node": "^14",
```
sp-watson, line 31: Hmm. for some reason the package-lock wasn't updated?

PR #255, opened 2022-06-14
NN-4164: Age of the prisoner page

natclamp-moj:


server/routes/ageOfPrisoner/ageOfPrisoner.ts
```diff
@@ -0,0 +1,78 @@
...
+    } catch (postError) {
+      logger.error(`Failed to post prison rule for draft adjudication: ${postError}`)
+      const idValue: number = parseInt(adjudicationNumber as string, 10)
+      res.locals.redirectUrl = adjudicationUrls.ageOfPrisoner.urls.start(idValue)
+      throw postError
+    }
+  }
+
+  getNextPageAfterCancel = (draftAdjudication: DraftAdjudication) => {
+    if (draftAdjudication.offenceDetails?.length) {
```
sp-watson, line 73: We don't want to duplicate this logic as it is pivotal to the flow. I notice there is already an instance of this in `placeOnReportService` BUT this is inconsistent with every other page isn't it? IIRC we agreed on a page navigation strategy that users would have to go thru all the pages once they entered the journey.

I think this means we have to have 2 versions of this page just like we have 2 versions of all the other pages (e.g.the statement page). The `nextPage` would be dependent on which page type it is - so if editing a new draft it goes to `/incident-role/{adjNo}` and if editing an existing draft it goes to `/incident-role/{adjNo}/submitted/edit` (I am not sure what the URLs are yet)

server/routes/ageOfPrisoner/ageOfPrisoner.ts
```diff
@@ -0,0 +1,78 @@
...
+      return adjudicationUrls.checkYourAnswers.urls.start(draftAdjudication.id)
```
sp-watson, line 74: Actually this is wrong anyway - but i guess we have to revisit it anyway?

sp-watson (approved):
Looks good - I am assuming we need to revisit the page wiring anyway

PR #263, opened 2022-06-20
NN-4251: GA CSP bug

natclamp-moj:


server/middleware/setUpWebSecurity.ts
```diff
@@ -23,14 +23,14 @@ export default function setUpWebSecurity(): Router {
             "'self'",
             (req, res: Response) => `'nonce-${res.locals.cspNonce}'`,
             'code.jquery.com',
-            'www.googletagmanager.com',
-            'www.google-analytics.com',
+            'https://www.googletagmanager.com',
```
sp-watson, line 26: I don't think this will work (well, CVL don't set these). Not sure why but best to only change what we have to

sp-watson (approved):

PR #268, opened 2022-06-21
NN-4175

timoj58:


server/services/placeOnReportService.ts
```diff
@@ -284,7 +284,11 @@ export default class PlaceOnReportService {
       handoverDeadline: draftAdjudication.incidentDetails.handoverDeadline,
       statementPresent: !!draftAdjudication.incidentStatement,
       statementComplete: draftAdjudication.incidentStatement?.completed || false,
-      offenceDetailsComplete: draftAdjudication.offenceDetails?.length > 0,
+      offenceDetailsComplete:
+        draftAdjudication.offenceDetails?.length > 0 &&
+        draftAdjudication.isYouthOffender !== null &&
```
sp-watson, line 289: I don't see the need for this change - it's not that complicated!

Users can't add offences without the other stuff set - so if they have an offence then they must have set the role and the rule. And ultimately the ticket says "COMPLETED - user has selected ‘Save and continue’ on the Offence Details page", which means that at least 1 offence must exist

timoj58: ok can remove, i added it as all three are needed.  Happy to let front end just use existing

sp-watson (approved):

PR #276, opened 2022-06-22
NN-4262

timoj58:


server/routes/detailsOfOffence/detailsOfOffence.ts
```diff
@@ -21,11 +21,11 @@ export default class DetailsOfOffenceRoutes {
     const { draftAdjudication, incidentRole, prisoner, associatedPrisoner } =
       await this.decisionTreeService.draftAdjudicationIncidentData(adjudicationNumber, user)
     const allOffences = await this.helper.populateSessionIfEmpty(adjudicationNumber, req, res)
+    const isYouthOffender = draftAdjudication.isYouthOffender || false
```
sp-watson, line 24: This class and the related njk really need an overhaul as it does lots of things it doesn't need to when there are no offences.

I'll make another ticket to sort this out because it can be made far simpler

sp-watson (approved):

server/routes/detailsOfOffence/detailsOfOffence.ts
timoj58: sure.  i just thought i would keep it consistent, it was using the null value lower down.   why i moved this line so it was in scope

PR #291, opened 2022-06-30
NN-4097: Fix confusion around session and API/draft offences

sp-watson:


integration_tests/integration/detailsOfOffence.cy.ts
```diff
@@ -292,30 +293,35 @@ context('Details of offence', () => {
...
+          offenceCode: 4001,
         },
       ],
     }).then((val: request.Response) => {
       expect(JSON.parse(val.text).count).to.equal(1)
     })
   })
 
-  it('goes to the reported adjudication statement edit page', () => {
+  it.only('goes to the reported adjudication statement edit page', () => {
```
natclamp-moj, line 324: might need to take the .only out! :)

PR #304, opened 2022-07-21
NN-4353: Add placeholder for prisoner outside establishment questions/answers

sp-watson:


server/services/decisionTreeService.ts
```diff
@@ -111,17 +111,26 @@ export default class DecisionTreeService {
...
       victimPrisoner,
       victimStaff,
+      victimPrisonerNumber: answerData.victimPrisonersNumber,
     }
   }
 
+  private async getPrisonerDetails(answerData: AnswerData, user: User): Promise<PrisonerResultSummary> {
+    // The victimOtherPersonsName will be set if the prisoner is from outside the establishment,
+    // in which case we do not want to get the extra prisoner information
+    if (!answerData.victimPrisonersNumber || answerData.victimOtherPersonsName) {
```
sp-watson, line 128: This is not nice, but the alternatives involve quite a lot of work.

I have taken a pragmatic approach - make it work with what we have and add a comment and some tests.

natclamp-moj (approved):

PR #311, opened 2022-07-25
NN-4358: Go to offence details page from incident role page if offences exist

sp-watson:


integration_tests/integration/incidentRoleSubmittedEdit.cy.ts
```diff
@@ -31,6 +31,16 @@ context('Incident role (edit after completion of report)', () => {
             dateTimeOfIncident: '2021-11-03T13:10:00',
             locationId: 27029,
           },
+          offenceDetails: [
```
sp-watson, line 34: There should always be offence details when editing submitted adjudications

timoj58 (approved):

PR #316, opened 2022-08-01
NN-4383: Update confirmation page

natclamp-moj:


server/routes/confirmedOnReport/confirmedOnReportChangeReport.ts
```diff
@@ -27,16 +27,10 @@ export default class ConfirmedOnReportChangeReportRoutes {
     )
 
     return res.render(`pages/confirmedOnReport`, {
```
sp-watson, line 29: Doesn't all this deleting mean that the above code isn't needed any more?

Maybe even `reportedAdjudicationsService .getConfirmationDetailsChangedReport` isn't required?

server/routes/confirmedOnReport/confirmedOnReportChangeReport.test.ts
```diff
@@ -38,8 +38,7 @@ describe('GET /prisoner-placed-on-report - edited adjudication', () => {
       .expect('Content-Type', /html/)
       .expect(res => {
         expect(res.text).toContain('John Smith’s report has been changed')
-        expect(res.text).toContain('You must follow local processes.')
```
sp-watson, line 41: as per the ts file comment - we must be able to get rid of lots of mocking for this test now?

server/routes/confirmedOnReport/confirmedOnReport.ts
```diff
@@ -19,22 +19,10 @@ export default class ConfirmedOnReportRoutes {
     )
 
     return res.render(`pages/confirmedOnReport`, {
-      adjudicationNumber,
```
sp-watson, line 22: If all this is deleted then can we not get rid of lots of the code above, maybe even delete `reportedAdjudicationsService.getConfirmationDetails`?

server/routes/confirmedOnReport/confirmedOnReport.test.ts
```diff
@@ -51,64 +51,7 @@ describe('GET /prisoner-placed-on-report', () => {
       .expect(res => {
         expect(res.text).toContain('Your report for John Smith has been submitted for review')
         expect(res.text).toContain('123')
-        expect(res.text).toContain('John Smith’s preferred language is')
```
sp-watson, line 54: As per ts file comment - doesn't all this deleting mean we can get rid of the mocks?

sp-watson (approved):

PR #318, opened 2022-08-03
NN-4378: Check answers review updates

natclamp-moj:


server/services/reportedAdjudicationsService.ts
```diff
@@ -36,6 +37,70 @@ export default class ReportedAdjudicationsService {
...
+    const { reportedAdjudication } = adjudicationData
+
+    if (reportedAdjudication.status === ReportedAdjudicationStatus.AWAITING_REVIEW)
+      return { reviewStatus: reportedAdjudicationStatusDisplayName(reportedAdjudication.status) }
+
+    const reviewingOfficer =
+      reportedAdjudication.reviewedByUserId &&
+      (await this.hmppsAuthClient.getUserFromUsername(reportedAdjudication.reviewedByUserId, user.token))
+
+    const getReasonTitle = (status: ReportedAdjudicationStatus) => {
```
sp-watson, line 51: All the code below this ought not to be in a service as it's completely UI-related. Probably the best place for all this UI-related code is the njk file (or the nunjucksSetup maybe). 

But it's not worth it right now as i suspect there is a lot of UI-specific code already in the services.

sp-watson (approved):

PR #319, opened 2022-08-04
NN-4377: Include new review status section and update tests

natclamp-moj:


integration_tests/integration/prisonerReportReview.cy.ts
```diff
@@ -217,12 +230,79 @@ context('Prisoner report - reviewer view', () => {
...
 
+    PrisonerReportPage.reviewSummaryTitle().should('exist')
+    PrisonerReportPage.reviewSummary().should('not.exist')
     PrisonerReportPage.incidentDetailsSummary().should('exist')
     PrisonerReportPage.offenceDetailsSummary().should('exist')
     PrisonerReportPage.incidentStatement().should('exist')
     PrisonerReportPage.reportNumber().should('exist')
     PrisonerReportPage.reviewerPanel().should('exist')
   })
+  it('should contain the correct review summary - returned', () => {
```
sp-watson, line 241: It's good that these tests are here but 2 things make it a maintenance problem:

 - All this is currently tested in the jest tests. That should be enough - adding more tests means more tests to change. You do need 1 int test to cover the review data, but that should suffice.
 - It's inconsistent with prisonerReport.cy.ts. i would expect both to have the same tests, or neither, or share a test somehow.

integration_tests/integration/prisonerReport.cy.ts
```diff
@@ -401,4 +386,47 @@ context('Prisoner report - reporter view', () => {
       })
     })
   })
+  describe('review statuses', () => {
```
sp-watson, line 389: It's confusing that these tests are virtually the same as the above ones but these are for `review statuses` and the other is for `test prisoners`. In all they do test all 4 statuses but it's not obvious.

The problem is that if/when these come to be modified people will just not know what it tests and therefore what the values should be.

I have made a separate comment that there isn't much point checking these statuses in int tests as they are already covered by the jest tests.

sp-watson (approved):

PR #321, opened 2022-08-09
NN-4351

timoj58:
new pages for incite / assist


server/routes/associatedPrisoner/associatePrisonerPage.ts
```diff
@@ -0,0 +1,237 @@
...
+    if (fromSearch) {
+      roleAssociatedPrisoner = {
+        prisonerNumber: roleAssociatedPrisoner.prisonerNumber,
+        currentAssociatedPrisonerNumber: fromSearch,
+      }
+    } else if (fromDelete) {
+      roleAssociatedPrisoner = {
+        prisonerNumber: roleAssociatedPrisoner.prisonerNumber,
+      }
+    }
```
natclamp-moj, line 66: Could we do this better by having a separate function with these if statements in, and then we don't need to use 'let' and reassign?

server/routes/associatedPrisoner/associatePrisonerPage.ts
```diff
@@ -0,0 +1,237 @@
...
+      currentAssociatedPrisonerNumber: associatedPrisonersNumber,
+    }
+
+    const validationError = validateForm({
+      // eslint-disable-next-line no-nested-ternary
+      location: selectedAnswerId
+        ? selectedAnswerId === 'internal'
+          ? AssociatedPrisonerLocation.INTERNAL
+          : AssociatedPrisonerLocation.EXTERNAL
+        : AssociatedPrisonerLocation.UNKNOWN,
```
natclamp-moj, line 111: Could we use a switch statement here or something else so we don't have to use a nested ternary?

server/routes/associatedPrisoner/associatePrisonerPage.ts
```diff
@@ -0,0 +1,237 @@
...
+      case AssociatedPrisonerLocation.INTERNAL:
+        associatedPrisonersName = roleAssociatedPrisoner.currentAssociatedPrisonerNumber
+          ? await this.getCurrentAssociatedPrisonersName(
+              roleAssociatedPrisoner.currentAssociatedPrisonerNumber,
+              currentUser
+            )
+          : null
+        break
+      default:
+    }
```
natclamp-moj, line 216: Could we put this in a separate function instead of using `let` and then reassigning?

natclamp-moj (approved):

PR #342, opened 2022-10-03
NN-4523: Centralise pageType conditionals with helper

natclamp-moj:
I've created a centralised abstract class inside a helper file, which can then be accessed by damages, evidence and witness controllers so the logic is shared and each of the files is a bit tidier.


neilmendum (approved):
Nice!

PR #382, opened 2022-11-03
NN-4470: Schedule hearing

natclamp-moj:
Completes https://dsdmoj.atlassian.net/browse/NN-4470
- resolves bug in hearingDetails page where the moment format was incorrect, rendering the minutes as the current month!
- merges in date of discovery work
- turns on the hearing work flag in dev (will now render the tabbed version of the prisoner report page)


server/data/DraftAdjudicationResult.ts
```diff
@@ -155,7 +155,7 @@ export enum WitnessCode {
 }
 
 export type HearingDetails = {
-  id: number
+  id?: number
   locationId: number
```
timoj58, line 159: oicHearingType field is now available on dev server and is mandatory so should really add under this PR, and default to GOV until other changes are made?

server/routes/adjudicationTabbedParent/scheduleHearing/scheduleHearingPage.ts
```diff
@@ -0,0 +1,190 @@
...
+  const values = {
+    hearingDetails: {
+      hearingDate: req.body.hearingDate,
+      locationId: req.body.locationId,
+    },
+  }
+  return values
+}
+
+const getHearingDate = (userProvidedValue?: SubmittedDateTime) => {
```
timoj58, line 181: i noticed this is also in incidentDetails page.  Would it better to move all instances of this to a utility class / function?

timoj58 (approved):

PR #383, opened 2022-11-04
NN-4500 all local unit/int tests pass

HamHazJustice:
Check your answers - added Date / Time of discovery

### DIFF
--- a/server/services/placeOnReportService.ts
+++ b/server/services/placeOnReportService.ts
@@ -159,11 +159,11 @@ export default class PlaceOnReportService {
         value: getFormattedOfficerName(reporter.name),
       },
       {
-        label: 'Date of Incident',
+        label: 'Date of incident',
         value: date,
       },
       {
-        label: 'Time of Incident',
+        label: 'Time of incident',
         value: time,
       },
       {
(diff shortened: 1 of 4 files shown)


server/services/placeOnReportService.ts
```diff
@@ -156,17 +159,25 @@ export default class PlaceOnReportService {
         value: getFormattedOfficerName(reporter.name),
       },
       {
-        label: 'Date',
+        label: 'Date of Incident',
```
timoj58, line 162: incident should be lower looking at the wireframe.  Design wise i think Anne always prefers upper case on first word of sentence only

server/services/placeOnReportService.ts
```diff
@@ -156,17 +159,25 @@ export default class PlaceOnReportService {
...
         value: date,
       },
       {
-        label: 'Time',
+        label: 'Time of Incident',
```
timoj58, line 166: ditto here lower case incident

server/services/placeOnReportService.ts
natclamp-moj: Yep Tim is correct

HamHazJustice: addressed throughout

server/services/placeOnReportService.ts
HamHazJustice: addressed throughout

natclamp-moj (approved):
Looks good

PR #400, opened 2022-11-17
NN-4461  - AllCompletedReports / YourComletedReports - Replace Select With CheckBox's

HamHazJustice:
3 unresolved issues (use of any and 2 unit test skip)

### DIFF
--- a/integration_tests/mockApis/adjudications.ts
+++ b/integration_tests/mockApis/adjudications.ts
@@ -3,7 +3,7 @@ import moment from 'moment'
 import { stubFor, verifyRequest } from './wiremock'
 import { apiPageResponseFrom } from '../../server/test/mojPaginationUtils'
 import { OffenceDetails } from '../../server/data/DraftAdjudicationResult'
-import { ReportedAdjudicationStatus } from '../../server/data/ReportedAdjudicationResult'
+import { allStatuses, ReportedAdjudicationStatus } from '../../server/data/ReportedAdjudicationResult'
 
 const stubPing = (status = 200): SuperAgentRequest =>
   stubFor({
@@ -231,7 +231,7 @@ const stubGetReportedAdjudications =
       `${prefix}agency/${agencyId}?page=${number}&size=${size}` +
       `${(filter.fromDate && `&startDate=${filter.fromDate}`) || ''}` +
       `${(filter.toDate && `&endDate=${filter.toDate}`) || ''}` +
-      `${(filter.status && `&status=${filter.status}`) || ''}`
+      `${(filter.status && `&status=${filter.status}`) || `&status=${allStatuses}`}` // TODO
     return stubFor({
       request: {
         method: 'GET',
--- a/integration_tests/mockApis/adjudications.ts
+++ b/integration_tests/mockApis/adjudications.ts
@@ -231,7 +231,7 @@ const stubGetReportedAdjudications =
       `${prefix}agency/${agencyId}?page=${number}&size=${size}` +
       `${(filter.fromDate && `&startDate=${filter.fromDate}`) || ''}` +
       `${(filter.toDate && `&endDate=${filter.toDate}`) || ''}` +
-      `${(filter.status && `&status=${filter.status}`) || `&status=${allStatuses}`}` // TODO
+      `${(filter.status && `&status=${filter.status}`) || `&status=${allStatuses}`}`
     return stubFor({
       request: {
         method: 'GET',
--- a/server/data/ReportedAdjudicationResult.ts
+++ b/server/data/ReportedAdjudicationResult.ts
@@ -48,7 +48,7 @@ export interface ReportedAdjudicationEnhanced extends ReportedAdjudication {
 export type ReportedAdjudicationFilter = {
   fromDate: moment.Moment
   toDate: moment.Moment
-  status: ReportedAdjudicationStatus[]
+  status: ReportedAdjudicationStatus | ReportedAdjudicationStatus[]
 }
 
 export type ScheduledHearingFilter = {
--- a/server/data/ReportedAdjudicationResult.ts
+++ b/server/data/ReportedAdjudicationResult.ts
@@ -48,7 +48,7 @@ export interface ReportedAdjudicationEnhanced extends ReportedAdjudication {
 export type ReportedAdjudicationFilter = {
   fromDate: moment.Moment
   toDate: moment.Moment
-  status?: ReportedAdjudicationStatus
+  status: ReportedAdjudicationStatus[]
 }
 
 export type ScheduledHearingFilter = {
--- a/server/data/ReportedAdjudicationResult.ts
+++ b/server/data/ReportedAdjudicationResult.ts
@@ -91,6 +91,14 @@ export enum ReportedAdjudicationStatus {
   ACCEPTED = 'ACCEPTED',
 }
 
+export const allStatuses = [
+  ReportedAdjudicationStatus.SCHEDULED,
+  ReportedAdjudicationStatus.UNSCHEDULED,
+  ReportedAdjudicationStatus.AWAITING_REVIEW,
+  ReportedAdjudicationStatus.RETURNED,
+  ReportedAdjudicationStatus.REJECTED,
+]
+
 export enum OicHearingType {
   GOV_ADULT = 'GOV_ADULT',
   GOV_YOI = 'GOV_YOI',
--- a/server/data/ReportedAdjudicationResult.ts
+++ b/server/data/ReportedAdjudicationResult.ts
@@ -84,11 +84,11 @@ export type ReviewAdjudication = {
 // This enum needs to be kept in sync with the enum in the API.
 export enum ReportedAdjudicationStatus {
   AWAITING_REVIEW = 'AWAITING_REVIEW',
+  UNSCHEDULED = 'UNSCHEDULED',
   RETURNED = 'RETURNED',
+  SCHEDULED = 'SCHEDULED',
   REJECTED = 'REJECTED',
   ACCEPTED = 'ACCEPTED',
-  SCHEDULED = 'SCHEDULED',
-  UNSCHEDULED = 'UNSCHEDULED',
 }
 
 export enum OicHearingType {
--- a/server/routes/allCompletedReports/allCompletedReports.test.ts
+++ b/server/routes/allCompletedReports/allCompletedReports.test.ts
@@ -85,7 +85,7 @@ afterEach(() => {
   jest.resetAllMocks()
 })
 
-describe('GET /all-completed-reports', () => {
+describe.skip('GET /all-completed-reports', () => {
   it('should render the not found page without the correct role', () => {
     userService.getUserRoles.mockResolvedValue([])
     return request(app)
@@ -95,7 +95,7 @@ describe('GET /all-completed-reports', () => {
         expect(res.text).not.toContain('All completed reports')
       })
   })
-  it.skip('should render the correct page with the correct role', () => {
+  it('should render the correct page with the correct role', () => {
     userService.getUserRoles.mockResolvedValue(['ADJUDICATIONS_REVIEWER'])
 
     return request(app)
@@ -141,7 +141,7 @@ describe('POST /all-completed-reports', () => {
     return request(app)
       .post(adjudicationUrls.allCompletedReports.root)
       .expect(res => {
-        expect(res.text).not.toContain('All completed reports')
+        expect(res.text).toContain('Page not found')
       })
   })
 })
--- a/server/routes/allCompletedReports/allCompletedReports.test.ts
+++ b/server/routes/allCompletedReports/allCompletedReports.test.ts
@@ -92,7 +92,7 @@ describe.skip('GET /all-completed-reports', () => {
       .get(adjudicationUrls.allCompletedReports.root)
       .expect('Content-Type', /html/)
       .expect(res => {
-        expect(res.text).not.toContain('All completed reports')
+        expect(res.text).toContain('Page not found')
       })
   })
   it('should render the correct page with the correct role', () => {
@@ -102,7 +102,7 @@ describe.skip('GET /all-completed-reports', () => {
       .get(adjudicationUrls.allCompletedReports.root)
       .expect('Content-Type', /html/)
       .expect(res => {
-        expect(res.text).toContain('All completed reports')
+        expect(res.text).toContain('Adjudications')
       })
   })
   it('should load the correct details', () => {
--- a/server/routes/allCompletedReports/allCompletedReports.test.ts
+++ b/server/routes/allCompletedReports/allCompletedReports.test.ts
@@ -85,7 +85,7 @@ afterEach(() => {
   jest.resetAllMocks()
 })
 
-describe.skip('GET /all-completed-reports', () => {
+describe('GET /all-completed-reports', () => {
   it('should render the not found page without the correct role', () => {
     userService.getUserRoles.mockResolvedValue([])
     return request(app)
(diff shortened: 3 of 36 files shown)


server/routes/allCompletedReports/allCompletedReports.test.ts
```diff
@@ -95,7 +95,7 @@ describe('GET /all-completed-reports', () => {
         expect(res.text).not.toContain('All completed reports')
       })
   })
-  it('should render the correct page with the correct role', () => {
+  it.skip('should render the correct page with the correct role', () => {
```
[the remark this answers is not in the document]
natclamp-moj: Also the test above this one checks for "all completed reports" - this should be changed to "Adjudications" as otherwise that one is a bit useless now - it'll never find "All completed reports"! :)

server/views/macros/adjudicationFilter.njk
```diff
@@ -2,10 +2,21 @@
 {% from "govuk/components/input/macro.njk" import govukInput %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
 {% from "govuk/components/select/macro.njk" import govukSelect %}
+{% from "govuk/components/checkboxes/macro.njk" import govukCheckboxes %}
 
 {% macro adjudicationFilter(filter, statuses, errors, clearUrl, csrfToken) %}
+  {% set statusInRequest = filter.status.toString() %}
+  {% set statusInRequestArr = statusInRequest.split(',')%}
```
natclamp-moj, line 9: I think these two lines of logic would be better done in the controller file (so you know exactly what you're passing as the filter). This will keep the template cleaner too, it's much easier to debug when it's on the server side

server/routes/allCompletedReports/allCompletedReports.test.ts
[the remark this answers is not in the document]
natclamp-moj: Perhaps change the first test to
```
it('should render the not found page without the correct role', () => {
    userService.getUserRoles.mockResolvedValue([])
    return request(app)
      .get(adjudicationUrls.allCompletedReports.root)
      .expect('Content-Type', /html/)
      .expect(res => {
        expect(res.text).toContain('Page not found')
      })
  })
```
and second one to
```
  it('should render the correct page with the correct role', () => {
    userService.getUserRoles.mockResolvedValue(['ADJUDICATIONS_REVIEWER'])

    return request(app)
      .get(adjudicationUrls.allCompletedReports.root)
      .expect('Content-Type', /html/)
      .expect(res => {
        expect(res.text).toContain('Date of discovery')
      })
  })
```

server/routes/yourCompletedReports/yourCompletedReport.test.ts
```diff
@@ -85,7 +85,7 @@ describe('GET /your-completed-reports', () => {
       })
     })
 
-    it('should load the your completed reported page', () => {
+    it.skip('should load the your completed reported page', () => {
```
natclamp-moj, line 88: I think if you make the changes I suggested to you on Slack, these two skipped tests will resolve themselves

server/routes/allCompletedReports/allCompletedReports.ts
```diff
@@ -43,6 +47,16 @@ export default class AllCompletedReportsRoutes {
...
+      if (uiFilter.status === undefined) {
+        /* eslint-disable @typescript-eslint/no-explicit-any */
+        uiFilter.status = [
+          ReportedAdjudicationStatus.AWAITING_REVIEW,
+          ReportedAdjudicationStatus.UNSCHEDULED,
+          ReportedAdjudicationStatus.RETURNED,
+          ReportedAdjudicationStatus.SCHEDULED,
+          ReportedAdjudicationStatus.REJECTED,
+        ] as any
+      }
```
natclamp-moj, line 59: I think this default array needs to be in the fillInDefaults function as I suggested on slack

server/views/macros/adjudicationFilter.njk
```diff
@@ -2,10 +2,21 @@
...
+  {% set items = [] %}
+  {% for status in statuses %}
+    {% set item = {text:status.text, value:status.value, checked: statusInRequestArr.includes(status.value) } %}
```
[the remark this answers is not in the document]
HamHazJustice: I think we should be ok as includes() is case sensitive and in our case the two schedule strings are not identical as the case is different - we can discuss - happy to come up with alternative if needed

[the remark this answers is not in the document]
natclamp-moj: Aren't we sending them through as all caps because we're using the enum?

server/routes/allCompletedReports/allCompletedReports.ts
```diff
@@ -46,7 +46,9 @@ export default class AllCompletedReportsRoutes {
       const filter = filterFromUiFilter(uiFilter)
       const results = await this.reportedAdjudicationsService.getAllCompletedAdjudications(
         res.locals.user,
-        filter,
+        /* eslint-disable */
+        filter as any, // TODO - NN-4461
+        /* eslint-enable */
```
natclamp-moj, line 51: We really shouldn't be disabling eslint for this - it's there for a reason! The only one we should use is the multiple classes per file one I think
Having to cast `filter` as any means the type needs updating somewhere else

server/routes/allCompletedReports/allCompletedReports.ts
```diff
@@ -31,7 +31,7 @@ export default class AllCompletedReportsRoutes {
   ): Promise<void> =>
     res.render(`pages/allCompletedReports`, {
       allCompletedReports: results,
-      filter,
+      filter: filter == null ? [{ status: 'SCHEDULED' }] : filter,
```
natclamp-moj, line 34: I thought that if no filter was provided it meant that all statuses should be selected? Not just SCHEDULED

server/routes/allCompletedReports/allCompletedReports.ts
natclamp-moj: I think you need to change the type in this function `FilterFromUiFilter` in the helper file
`status: filter.status || allStatuses`

server/routes/yourCompletedReports/yourCompletedReports.ts
```diff
@@ -41,7 +41,9 @@ export default class YourCompletedReportsRoutes {
     const filter = filterFromUiFilter(uiFilter)
     const results = await this.reportedAdjudicationsService.getYourCompletedAdjudications(
       res.locals.user,
-      filter,
+      /* eslint-disable */
+      filter as any, // TODOO - NN-4461
+      /* eslint-enable */
```
natclamp-moj, line 46: As with the other comment, we shouldn't disable eslint for this - the type needs changing somewhere - I think it will be sorted if you change the type I suggested above

server/views/macros/adjudicationFilter.njk
natclamp-moj: Sorry I'm really not a fan of theses two lines of logic in the njk file - I also don't see why it's necessary as aren't you passing them through as the enum, so they'd be represented as strings in the array anyway? 
Can you not just do the `filter.status.includes(status.value)`?

HamHazJustice: yub will do 👍 -)

integration_tests/mockApis/adjudications.ts
```diff
@@ -231,7 +231,7 @@ const stubGetReportedAdjudications =
       `${prefix}agency/${agencyId}?page=${number}&size=${size}` +
       `${(filter.fromDate && `&startDate=${filter.fromDate}`) || ''}` +
       `${(filter.toDate && `&endDate=${filter.toDate}`) || ''}` +
-      `${(filter.status && `&status=${filter.status}`) || ''}`
+      `${(filter.status && `&status=${filter.status}`) || `&status=${allStatuses}`}` // TODO
```
natclamp-moj, line 234: is there anything still `TODO` here

server/views/macros/adjudicationFilter.njk
```diff
@@ -5,16 +5,16 @@
...
   {% set items = [] %}
   {% for status in statuses %}
-    {% set item = {text:status.text, value:status.value, checked: statusInRequestArr.includes(status.value) } %}
+    {% set item = {text:status.text, value:status.value, checked: filter.status.includes(status.value) } %}
       {% if status.value!=='ACCEPTED' %}
         {% set items = (items.push(item),items) %}
       {% endif%}
   {% endfor %}
 
+filter={{filter.status}}
```
natclamp-moj, line 16: Not sure this is meant to be here!

server/routes/allCompletedReports/allCompletedReports.ts
```diff
@@ -31,7 +31,7 @@ export default class AllCompletedReportsRoutes {
   ): Promise<void> =>
     res.render(`pages/allCompletedReports`, {
       allCompletedReports: results,
-      filter: filter == null ? [{ status: 'SCHEDULED' }] : filter,
+      filter: filter === null ? [{ status: `allStatuses` }] : filter,
```
natclamp-moj, line 34: You can just pass `filter` in here. No need for the ternary. This is because filter should never be equal to `null` - if you look at the functions involved this this line `const uiFilter = fillInDefaults(uiFilterFromRequest(req))`, it pulls the dates and statuses from the request, and if there's nothing there, default dates are passed in and the array of all the statuses.
You can test this by removing the ternary here and going to `/all-completed-reports` - all the statuses are checked even though there's no query parameters

server/utils/adjudicationFilterHelper.ts
```diff
@@ -35,26 +35,33 @@ export const uiFilterFromRequest = (req: Request): UiFilter => {
 // get results for all of today, thus if set the 'to' date 2 days ago we should also get the proceeding 2 days.
 // The status default to null, which the api interprets as all statuses.
 export const fillInDefaults = (uiFilter: UiFilter): UiFilter => {
+  /* const allStatuses = [
+    ReportedAdjudicationStatus.SCHEDULED,
+    ReportedAdjudicationStatus.UNSCHEDULED,
+    ReportedAdjudicationStatus.AWAITING_REVIEW,
+    ReportedAdjudicationStatus.RETURNED,
+    ReportedAdjudicationStatus.REJECTED,
+  ] */
```
natclamp-moj, line 44: I'm assuming we don't need these anymore seen as they're commented out?

server/data/ReportedAdjudicationResult.ts
```diff
@@ -48,7 +48,7 @@ export interface ReportedAdjudicationEnhanced extends ReportedAdjudication {
 export type ReportedAdjudicationFilter = {
   fromDate: moment.Moment
   toDate: moment.Moment
-  status?: ReportedAdjudicationStatus
+  status: ReportedAdjudicationStatus[]
```
natclamp-moj, line 51: I think this needs to be 
`status: ReportedAdjudicationStatus | ReportedAdjudicationStatus[]`
I've been through and logged the statuses, trying different combinations - when there's only one status sent, it's not sent as an array but a string. So the two alternatives are needed here

server/utils/adjudicationFilterHelper.ts
```diff
@@ -19,14 +19,14 @@ const error: { [key in ErrorType]: FormError } = {
 export type UiFilter = {
   fromDate: string
   toDate: string
-  status: ReportedAdjudicationStatus
+  status: ReportedAdjudicationStatus[]
```
natclamp-moj, line 22: I think this should be
`status: ReportedAdjudicationStatus | ReportedAdjudicationStatus[]`
Explained in other comment

server/utils/adjudicationFilterHelper.ts
```diff
@@ -19,14 +19,14 @@ const error: { [key in ErrorType]: FormError } = {
...
 }
 
 export const uiFilterFromRequest = (req: Request): UiFilter => {
   return {
     fromDate: req.query.fromDate as string,
     toDate: req.query.toDate as string,
-    status: req.query.status as ReportedAdjudicationStatus,
+    status: req.query.status as ReportedAdjudicationStatus[],
```
natclamp-moj, line 29: This needs to be `status: req.query.status as ReportedAdjudicationStatus`
As I don't think the query value is ever an array?

server/services/reportedAdjudicationsService.test.ts
```diff
@@ -149,7 +149,7 @@ describe('reportedAdjudicationsService', () => {
         {
           fromDate: moment('01/01/2021', 'DD/MM/YYYY'),
           toDate: moment('01/09/2021', 'DD/MM/YYYY'),
-          status: ReportedAdjudicationStatus.AWAITING_REVIEW,
+          status: [ReportedAdjudicationStatus.AWAITING_REVIEW],
```
natclamp-moj, line 152: As it's just the one status you're passing in, it would have to be in an array, as the code that generates what's passed in as the status is 
`status: filter.status || allStatuses,` (in filterFromUIFilter function)

server/services/reportedAdjudicationsService.test.ts
```diff
@@ -400,7 +400,7 @@ describe('reportedAdjudicationsService', () => {
         {
           fromDate: moment('01/01/2021', 'DD/MM/YYYY'),
           toDate: moment('01/01/2021', 'DD/MM/YYYY'),
-          status: ReportedAdjudicationStatus.AWAITING_REVIEW,
+          status: [ReportedAdjudicationStatus.AWAITING_REVIEW],
```
natclamp-moj, line 403: As it's just the one status you're passing in, it would have to be in an array, as the code that generates what's passed in as the status is 
`status: filter.status || allStatuses,` (in filterFromUIFilter function)

integration_tests/mockApis/adjudications.ts
HamHazJustice: no nothing TODO - removed

server/views/macros/adjudicationFilter.njk
HamHazJustice: GOOD SPOT - removed

server/data/ReportedAdjudicationResult.ts
HamHazJustice: I think all works with the status's as an array , but this may explain why I still had a TODO on the line
`${(filter.status && `&status=${filter.status}`) || `&status=${allStatuses}`}` // TODO
As i above &status=${allStatuses}` is not an array 
Happy put your suggestion back in - perhaps we should revisit this at a later date.

server/utils/adjudicationFilterHelper.ts
HamHazJustice: So here is how the array appears in a query i just ran:

http://localhost:3000/all-completed-reports?fromDate=15%2F11%2F2022&toDate=22%2F11%2F2022&status=AWAITING_REVIEW&status=RETURNED&status=REJECTED

So the checkboxes when submitted produce essentially an array of status's
status=AWAITING_REVIEW&status=RETURNED&status=REJECTED

All the tests work using :
  status: ReportedAdjudicationStatus[]

So I dont think there is any need to make that change

server/utils/adjudicationFilterHelper.ts
HamHazJustice: See previous comment

server/services/reportedAdjudicationsService.test.ts
HamHazJustice: same as previous comments

server/services/reportedAdjudicationsService.test.ts
HamHazJustice: samme as previous comments

server/utils/adjudicationFilterHelper.ts
HamHazJustice: remobved

natclamp-moj (approved):

PR #403, opened 2022-11-21
Nn 4646 maintenance page

jamesgrant-moj:
As an organisation there are times when we will need to put the Adjudications service into a maintenance mode. As such we need a holding page we can enable and direct traffic to while the service is down.

DRAFT content below:

'This service is currently unavailable 

You'll be able to use the Adjudications service on DPS again from [9am on Monday, 23 May 2022]. [insert time/date in this format]

The page should be styled in line with the service but stand alone so we can take down all of the services and still use the page.

Technical implementation

Implemented in the Adjudications UI, rather than DPS, incase folks have bookmarked direct pages

Ideally should be behind a feature switch which is sourced from helm config

When true we should enable a catchall Express route which directs users to the maintenance page

### DIFF
--- a/server/app.ts
+++ b/server/app.ts
@@ -1,11 +1,11 @@
 import express from 'express'
-
 import path from 'path'
 import createError from 'http-errors'
 import config from './config'
 import GotenbergClient from './data/gotenbergClient'
 
 import indexRoutes from './routes'
+import maintenanceRoutes from './routes/maintenance-index'
 import nunjucksSetup from './utils/nunjucksSetup'
 import errorHandler from './errorHandler'
 import standardRouter from './routes/standardRouter'
@@ -20,6 +20,8 @@ import authorisationMiddleware from './middleware/authorisationMiddleware'
 import logger from '../logger'
 import { Services } from './services'
 import { pdfRenderer } from './utils/pdfRenderer'
+import 'dotenv/config'
+import maintenancePageRouter from './routes/maintenancePageRouter'
 
 export default function createApp(services: Services): express.Application {
   // We do not want the server to exit, partly because any log information will be lost.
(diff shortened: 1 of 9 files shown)


server/app.ts
```diff
@@ -46,7 +48,18 @@ export default function createApp(services: Services): express.Application {
   app.use(pdfRenderer(new GotenbergClient(config.apis.gotenberg.apiUrl)))
   app.use(authorisationMiddleware())
 
-  app.use('/', indexRoutes(standardRouter(services.userService), services))
+  if (process.env.DISPLAY_MAINTENANCE_PAGE === 'true') {
```
neilmendum, line 51: You could name this `MAINTENANCE_MODE` but I'm not too precious 😄

server/routes/maintenancePage/maintenancePage.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  app = appWithAllRoutes({ production: false }, { userService })
+  userService.getUserRoles.mockResolvedValue([])
+})
+
+afterEach(() => {
+  jest.resetAllMocks()
+})
+
+describe('GET /planned-maintenance', () => {
+  it('should get the maintenance page', () => {
```
neilmendum, line 23: Could be worth testing that for a different route we still land on the maintenance page?

server/views/pages/maintenancePage.njk
```diff
@@ -0,0 +1,33 @@
...
+  <div class="govuk-width-container">
+
+    {% set mainClasses = "govuk-main-wrapper--l" %}
+    <div class="govuk-grid-row govuk-!-margin-top-7 govuk-!-margin-bottom-7">
+      <div class="govuk-grid-column-two-thirds">
+        <h1 class="govuk-heading-l">This service is currently unavailable</h1>
+        <p class="govuk-body">
+          You'll be able to use the Adjudications service on DPS again from [9am on Monday, 23 May 2022]
+        </p>
+        <h2 class="govuk-heading-m">Help</>
```
neilmendum, line 22: Anne made a comment about a tweak to the copy - see [Jira](https://dsdmoj.atlassian.net/browse/NN-4646?focusedCommentId=252430)

server/app.ts
neilmendum: And I would add this (as `false`) to the `env` section of
- `helm_deploy/values-dev.yaml`
- `helm_deploy/values-preprod.yaml`
- `helm_deploy/values-prod.yaml`

neilmendum (commented):
Looks good - added a few comments

natclamp-moj (commented):
Just querying the tax credits paragraph with the team, not sure that's meant to be on the screenshot...
Make sure to change your PR title to NN-4646! Other than that, looks good if Neil's suggestions are done 👍

server/app.ts
jamesgrant-moj: Updated all these changes

neilmendum (approved):
LGTM

neilmendum (approved):

PR #417, opened 2022-11-30
NN-4618: select prisoner gender

natclamp-moj:
This PR completed NN-4618 and NN-4628
- creates a new page to gather the gender of the prisoner with a simple set of radio buttons
- the page is only shown to users when the gender of the prisoner has not been set on that prisoner's profile, which it does by checking the gender value on the select prisoner page. The link to start the adjudication is conditional on the value brought back on the prisoner object when the user conducts a search
- it adds this gender data to the session, where it is read from on the incident details (the following page) and submitted as part of the api call to begin a new adjudication
- if the gender was already specified on the prisoners profile, the path skips over the new page, and goes straight to incident details. On incident details, it will read the gender from the data achieved by calling prison api instead of checking the session
- There is an edit version of the new page which will be linked from the 'check your answers' page (ticket still to be completed). On this version of the page, it reads the draft adjudication and pre-selects the gender. The selected gender (whether it has been changed or not) is sent to the new adjudications api endpoint to change gender. The user is then sent back to the check your answers page.
- There's a lot of changed test files as gender is now being included on an important data structure - sorry!

Potential snags/questions: will the user be allowed to change the gender of the prisoner on the 'edit after submission' version of the check your answers page? If so, we might need to change some of the redirects to make sure that the user is sent back to the correct version of the check your answers page - but this can be done later


server/routes/selectGender/selectGenderPage.ts
```diff
@@ -0,0 +1,104 @@
...
+    })
+  }
+
+  view = async (req: Request, res: Response): Promise<void> => {
+    const draftId = Number(req.params.draftId) || null
+    const { user } = res.locals
+
+    let readApiGender: string = null
+    if (this.pageOptions.isEdit()) {
+      readApiGender = await this.getPreviouslySelectedGenderFromApi(draftId, user)
```
timoj58, line 54: re edit gender, i think its only available if the prisoner was undefined previously?

timoj58 (approved):

PR #420, opened 2022-12-01
NN-4676 - Adjudications - New View All Completed Reports tab- Bug Ticket

HamHazJustice:


server/views/pages/allCompletedReports.njk
```diff
@@ -50,15 +63,14 @@
...
+                  attributes: {
           "aria-sort": "descending"
         }
         },
-        { text: "Reporting officer" },
+        { html: "Name and prison number" },
         { text: "Status" },
+        { text: "Latest scheduled hearing" },
         { text: "" }
       ],
```
natclamp-moj, line 75: you need another `{ text: "" }` here for the final header space (otherwise the table header is too short

server/services/reportedAdjudicationsService.ts
```diff
@@ -298,6 +298,10 @@ export default class ReportedAdjudicationsService {
     const friendlyName =
       (prisonerResult && convertToTitleCase(`${prisonerResult.firstName} ${prisonerResult.lastName}`)) || ''
     const reportingOfficer = (reporterName && convertToTitleCase(reporterName)) || ''
+    const nextSheduledHearingDate =
```
natclamp-moj, line 301: Can we rename the variable something like `lastScheduledHearingDate` ? I think naming it `nextScheduledHearingDate` is confusing as it might not be the next one, it's the final one, isn't it?

server/services/reportedAdjudicationsService.ts
```diff
@@ -298,6 +298,10 @@ export default class ReportedAdjudicationsService {
...
+      reportedAdjudication.status === ReportedAdjudicationStatus.SCHEDULED
+        ? reportedAdjudication.hearings[reportedAdjudication.hearings.length - 1].dateTimeOfHearing
+        : ''
```
natclamp-moj, line 304: From the screenshot it looks like Simon wants there to be a hyphen if there's no scheduled hearing date to display here. So I think the best way to do this with your approach is to return `null` here if the status isn't SCHEDULED and then below (I will leave a comment where) you can use `|| "-"` as otherwise it will return an invalid date

server/services/reportedAdjudicationsService.ts
```diff
@@ -309,7 +313,13 @@ export default class ReportedAdjudicationsService {
         reportedAdjudication.incidentDetails.dateTimeOfIncident,
         'D MMMM YYYY - HH:mm'
       ),
+      dateTimeOfDiscovery: reportedAdjudication.incidentDetails.dateTimeOfDiscovery,
+      formattedDateTimeOfDiscovery: formatTimestampToDate(
+        reportedAdjudication.incidentDetails.dateTimeOfDiscovery,
+        'D MMMM YYYY - HH:mm'
+      ),
       statusDisplayName: reportedAdjudicationStatusDisplayName(reportedAdjudication.status),
+      formattedDateTimeOfScheduledHearing: formatTimestampToDate(nextSheduledHearingDate, 'D MMMM YYYY - HH:mm'),
```
natclamp-moj, line 322: You should be able to use 
`formatTimestampToDate(nextSheduledHearingDate, 'D MMMM YYYY - HH:mm') || "-",` here instead to show the hyphen when there's no hearing to display

natclamp-moj: I think this means you'll have to correct a few of the tests in the `reportedAdjudicationsService.test.ts` by the way, as it looks like in some of the expected data it returns an empty string but now it will returns "-"

integration_tests/integration/allCompletedReports.cy.ts
```diff
@@ -298,6 +302,131 @@ context('All Completed Reports', () => {
...
+    allCompletedReportsPage.paginationResults().should('have.text', 'Showing 21 to 40 of 300 results')
+    allCompletedReportsPage.resultsTable().should('exist')
+    allCompletedReportsPage
+      .resultsTable()
+      .find('td')
+      .then($data => {
+        expect($data.get(0).innerText).to.contain('15 November 2345 - 11:30')
+        expect($data.get(1).innerText).to.contain('Potter, Harry - A1234AA')
+        expect($data.get(2).innerText).to.contain('Scheduled')
+        expect($data.get(3).innerText).to.contain('30 November 2023 - 10:')
```
natclamp-moj, line 424: The time here is only `10:`, shouldn't it show the minutes too?

natclamp-moj (approved):
LGTM

PR #428, opened 2022-12-07
NN-4604 - Continue Report -updates to prisoner name and date columns and headings

HamHazJustice:


server/views/pages/continueReportSelect.njk
```diff
@@ -39,10 +36,9 @@
           "data-sort-value": report.displayName
         }
       },
-      { html: incidentDateHtml, attributes: {
+      { html: discoveryDateTimeHtml, attributes: {
           "data-sort-value": report.incidentDetails.dateTimeOfIncident
```
natclamp-moj, line 40: The data sort value needs changing here - otherwise when the table is resorted by the user, it'll sort on the incident dateTIme and not the discoveryDateTime

server/views/pages/continueReportSelect.njk
```diff
@@ -52,20 +48,19 @@
...
+            text: "Name and prison number",
             attributes: {
               "aria-sort": "ascending",
               "data-qa": "prisoner-name-sort"
             }
           },
-          { text: "Date of incident", 
+          { text: "Discovery date and time", 
             attributes: {
               "aria-sort": "ascending",
```
natclamp-moj, line 59: I think we need to check with Simon whether he wants this data sorting ascending or descending - it looks like in the screenshot he wants it descending (at the minute it's ascending)

HamHazJustice: well spotted!

server/views/pages/continueReportSelect.njk
HamHazJustice: good spot!

timoj58 (approved):

PR #434, opened 2022-12-09
NN-4709: Make link conditional

natclamp-moj:
This PR is to make the 'add date and time' link on the 'confirm DIS Forms have been issued' page is conditional, based on whether a user has already added a date already. It checks on the adjudication data to see if there's a date and returns a boolean, which is used to decide whether the link renders or not.
Also in this PR I tidied up the new test data class - specifically the completed adjudication function, so that we can add in whatever other data we might need in that returned dataset to be able to use it for the tests we're creating.


server/routes/testutils/testData.ts
```diff
@@ -1,24 +1,14 @@
...
-  dateTimeOfIssue: string
-  formattedDateTimeOfIssue: string
-}
-
 export default class TestData {
   completedAdjudication = (
     adjudicationNumber: number,
     prisonerNumber: string,
-    otherData?: OtherData,
+    otherData?: any,
```
natclamp-moj, line 11: Using `any` here because my plan is to be able to use this to insert various different types of data into the function in order to give the correct adjudication data for whatever test we need it for

timoj58 (approved):

PR #446, opened 2022-12-14
NN-4633: Print completed DIS1/2 forms

natclamp-moj:


server/data/prisonApiClient.ts
```diff
@@ -65,7 +66,18 @@ export default class PrisonApiClient {
...
       data: prisonerNumbers,
     })
-    return result.map(_ => plainToClass(PrisonerResult, _, { excludeExtraneousValues: true }))
+
+    return result.map(_ => plainToClass(PrisonerResult, _, { excludeExtraneousValues: false }))
+  }
+
+  async getAlertsForPrisoner(prisonerNumber: string): Promise<PrisonerAlerts> {
+    const alerts = await this.restClient.get<Alert[]>({
+      path: `/api/offenders/${prisonerNumber}/alerts/v2?alertCodes=PRGNT,HA,CSIP,PEEP`,
```
timoj58, line 75: minor but would it be better to use the values from alertFlagLabels, ie map out the codes, that way the two are in sync?  rather than hard code them here as well

timoj58 (approved):

PR #506, opened 2023-02-17
NN-4665

timoj58:
not proceed page

-  only create, as no update on server currently for outcome

### DIFF
--- a/server/views/pages/notProceed.njk
+++ b/server/views/pages/notProceed.njk
@@ -35,34 +35,34 @@
             classes: ''
         },
         items: [
-            {  value: "RELEASED",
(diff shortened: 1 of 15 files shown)


server/views/pages/notProceed.njk
```diff
@@ -0,0 +1,102 @@
...
+
+            {{ govukSelect({
+        id: 'notProceedReason',
+        name: 'notProceedReason',
+        label: {
+            text: 'Reason',
+            isPageHeading: false,
+            classes: ''
+        },
+        items: [
```
natclamp-moj, line 37: just wondering if we could we use an enum here instead of the strings?

natclamp-moj (approved):

PR #522, opened 2023-03-01
NN-4892: Fix enums, add int tests

natclamp-moj:
- add integration tests for completed versions of hearing details page
- fix witness enum
- fix 'OTHER' on adjourn enum
- add extra check so that 'not proceed' outcome table doesn't show on complete hearing -> not proceed route version

### DIFF
--- a/server/services/reportedAdjudicationsService.ts
+++ b/server/services/reportedAdjudicationsService.ts
@@ -794,7 +794,10 @@ export default class ReportedAdjudicationsService {
     if (!history.length || readOnly) return null
     const finalHistoryItem = history[history.length - 1]
     if (finalHistoryItem.outcome) {
-      if (finalHistoryItem.outcome.outcome.code === OutcomeCode.NOT_PROCEED) {
+      if (
+        finalHistoryItem.outcome.outcome.code === OutcomeCode.NOT_PROCEED &&
+        finalHistoryItem.hearing?.outcome.code !== HearingOutcomeCode.COMPLETE
+      ) {
         return {
           text: 'Remove this outcome',
           name: 'removeOutcomeButton',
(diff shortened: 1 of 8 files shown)


server/services/reportedAdjudicationsService.ts
```diff
@@ -794,7 +794,10 @@ export default class ReportedAdjudicationsService {
     if (!history.length || readOnly) return null
     const finalHistoryItem = history[history.length - 1]
     if (finalHistoryItem.outcome) {
-      if (finalHistoryItem.outcome.outcome.code === OutcomeCode.NOT_PROCEED) {
+      if (
+        finalHistoryItem.outcome.outcome.code === OutcomeCode.NOT_PROCEED &&
+        finalHistoryItem.hearing?.outcome.code !== HearingOutcomeCode.COMPLETE
```
natclamp-moj, line 799: Adding an extra check here so that the 'remove hearing' button shows when the hearing is complete and finding is NOT_PROCEED, otherwise it shows the the 'remove outcome' button

server/views/pages/adjudicationForReport/hearingTab.njk
```diff
@@ -32,7 +32,7 @@
         {% if item.hearing %}
           {% set hearingNumber = hearingNumber + 1 %}
           {{ hearingTable(hearingNumber, item.hearing, item.outcome, readOnly, reportNo, latestHearingId) }}
-          {% if item.outcome %} 
+          {% if item.outcome and item.hearing.outcome.code != HearingOutcomeCode.COMPLETE %}
```
natclamp-moj, line 35: extra check for when the hearing is complete and finding is NOT_PROCEED - without this, the 'not proceed' table will show under the hearing, which isn't what we want here (but is when the outcome is NOT_PROCEED in other circumstances)

timoj58 (approved):

PR #541, opened 2023-03-14
NN-4959: Adapt int tests and add function to get last outcome from array

natclamp-moj:
- changes 'next step' label to 'outcomes' label on summary
- adds change links for name of adjudicator and outcome data
- **does not add** any logic as to when these change links are visible - this will be done in a later PR
- adds new function to get adjudication and retrieve the final outcome item

### DIFF
--- a/server/views/partials/hearingDetailsVariations/hearingTableExpansions/hearingTableCompleteDismissed.njk
+++ b/server/views/partials/hearingDetailsVariations/hearingTableExpansions/hearingTableCompleteDismissed.njk
@@ -1,3 +1,24 @@
+{% set adjudicatorChangeLink = {
+  items: [
+    {
+      href: adjudicationUrls.enterHearingOutcome.urls.edit(adjudicationNumber, hearingId), 
+      text: "Change",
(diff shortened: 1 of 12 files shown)


server/views/partials/hearingDetailsVariations/hearingTableExpansions/hearingTableCompleteDismissed.njk
```diff
@@ -1,18 +1,41 @@
...
+  items: [
+    {
+      href: adjudicationUrls.enterHearingOutcome.urls.edit(adjudicationNumber, hearingId), 
+      text: "Change",
+      visuallyHiddenText: ' outcome',
+      attributes: {"data-qa": "change-link-hearing-outcome-outcome" }
+    }
+  ]
+} %}
+
```
natclamp-moj, line 21: I tried having this section as a partial so we could just include it in the multiple files where it's required but I'm not sure that 'set' registers when you use it like that - I might try again at a later date to see if i can get it to work another way

timoj58 (approved):

PR #644, opened 2023-06-02
NN-5110: Fix logging, move dotenv etc

natclamp-moj:
Introduces some changes explained in this PR: https://github.com/ministryofjustice/hmpps-template-typescript/pull/120


timoj58 (approved):

server.ts
```diff
@@ -1,3 +1,6 @@
+// Require app insights before anything else to allow for instrumentation of bunyan and express
+import 'applicationinsights'
```
andrewrlee, line 2: not sure this is strictly necessary but won't break anything

natclamp-moj: I did wonder why we didn't have it in the first place, but thought maybe it was something ~ mysterious ~ that happened once you imported haha!

PR #649, opened 2023-06-06
NN-4601: Update helmet with flag

natclamp-moj:
We noticed the helmet update meant that the stylesheet couldn't be loaded into the PDFs. After some research, we found that this was because in the new version, crossOriginResourcePolicy is set to true by default, whereas before it was false. 
The PDFs are generated by gotenburg on a different port, so none of the other options for the policy allow the stylesheet to load.


neilmendum (approved):
LGTM

neilmendum (approved):
LGTM

PR #694, opened 2023-07-03
Minor Helm tidyups

neilmendum:


helm_deploy/values-prod.yaml
```diff
@@ -2,7 +2,6 @@
 # Per environment values which override defaults in hmpps-manage-adjudications/values.yaml
 
 generic-service:
-  replicaCount: 4
```
neilmendum, line 5: Defined in `values.yaml`

PR #720, opened 2023-07-19
NN-5264: Refactor

natclamp-moj:


server/views/macros/activateSuspendedPunishmentsTable.njk
```diff
@@ -10,7 +10,7 @@
     {% set rows = (rows.push(
       [
           {
-            text: sanction.type | convertPunishmentType(sanction.stoppagePercentage, sanction.privilegeType, sanction.otherPrivilege),
+            text: sanction.type | convertPunishmentType(false, null, sanction.stoppagePercentage, sanction.privilegeType, sanction.otherPrivilege),
```
neilmendum, line 13: Do we need to change this `text` to `html`?

And also for other templates which use `convertPunishmentType`

neilmendum (approved):
LGTM

PR #724, opened 2023-07-20
NN-5338 allow session punishments to be removed

timoj58:


timoj58:
on second thoughts, this is not the right approach and perhaps we dont need to handle this case, or add a remove all punishments button instead?  UI would need to change to reflect current code as it doesnt look great

PR #739, opened 2023-07-26
NN-5289: fix by treating the flow as a create, rather than an edit

neilmendum:
https://dsdmoj.atlassian.net/browse/NN-5289

`manualEdit/redisId` was `undefined` because we're not updating an existing punishment

### DIFF
--- a/server/routes/additionalDays/numberOfAdditionalDays/numberOfAdditionalDaysPage.ts
+++ b/server/routes/additionalDays/numberOfAdditionalDays/numberOfAdditionalDaysPage.ts
@@ -106,7 +106,7 @@ export default class NumberOfAdditionalDaysPage {
       return adjudicationUrls.isPunishmentSuspended.urls.edit(adjudicationNumber, req.params.redisId)
     }
     if (this.pageOptions.isManualEdit()) {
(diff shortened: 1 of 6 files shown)


server/routes/additionalDays/numberOfAdditionalDays/numberOfAdditionalDaysPage.ts
```diff
@@ -106,7 +106,7 @@ export default class NumberOfAdditionalDaysPage {
       return adjudicationUrls.isPunishmentSuspended.urls.edit(adjudicationNumber, req.params.redisId)
     }
     if (this.pageOptions.isManualEdit()) {
-      return adjudicationUrls.whichPunishmentIsItConsecutiveToManual.urls.edit(adjudicationNumber, req.params.redisId)
+      return adjudicationUrls.whichPunishmentIsItConsecutiveToManual.urls.start(adjudicationNumber)
```
neilmendum, line 109: Assumed we'd want to switch to using `whichPunishmentIsItConsecutiveToManual.start` as `edit` goes into [this block](https://github.com/ministryofjustice/hmpps-manage-adjudications/blob/79ba2403cbcaaeb7a953630007501809fa950fff/server/routes/additionalDays/whichPunishmentConsecutiveTo/whichPunishmentConsecutiveToPage.ts#L76-L82)

PR #747, opened 2023-08-01
NN-5429: migrate calls to HMPPS Auth endpoints to use Manage Users API

neilmendum:
https://dsdmoj.atlassian.net/browse/NN-5429


integration_tests/mockApis/users.ts
```diff
@@ -0,0 +1,129 @@
+import { Response, SuperAgentRequest } from 'superagent'
+import { stubFor } from './wiremock'
+
+const stubUser = ({ username = 'USER1', activeCaseLoadId = 'MDI' }: { username?: string; activeCaseLoadId?: string }) =>
+  stubFor({
+    request: {
+      method: 'GET',
+      urlPattern: '/users/users/me',
```
neilmendum, line 8: In `feature.env` we use single worded paths, for example
`TOKEN_VERIFICATION_API_URL=http://localhost:9091/verification`
so I went with
`HMPPS_MANAGE_USERS_API_URL=http://localhost:9091/users`

Are folks ok with `urlPattern: '/users/users/...` ?

natclamp-moj (approved):

PR #760, opened 2023-08-04
NN-5449: store PunishmentDataV2 in Redis session and for API calls

neilmendum:
https://dsdmoj.atlassian.net/browse/NN-5449

### DIFF
--- a/server/data/PunishmentResult.ts
+++ b/server/data/PunishmentResult.ts
@@ -25,24 +25,6 @@ export enum PrivilegeType {
   OTHER = 'OTHER',
 }
 
-/**
- * @deprecated The method should not be used
- */
-export type PunishmentData = {
-  id?: number
-  redisId?: string
-  type: PunishmentType
-  privilegeType?: PrivilegeType
-  otherPrivilege?: string
-  stoppagePercentage?: number
-  days: number
-  startDate?: string
-  endDate?: string
-  suspendedUntil?: string
-  activatedFrom?: number
-  consecutiveReportNumber?: number
-}
-
 export type PunishmentDataV2 = {
   id?: number
   redisId?: string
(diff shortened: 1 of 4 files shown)


server/data/PunishmentResult.ts
```diff
@@ -196,7 +177,6 @@ export function flattenPunishment(punishment: PunishmentDataWithSchedule): Punis
     ...(endDate && { endDate }),
     ...(activatedFrom && { activatedFrom }),
     ...(consecutiveReportNumber && { consecutiveReportNumber }),
-    ...(consecutiveReportAvailable && { consecutiveReportAvailable }),
```
neilmendum, line 199: This was not present on `PunishmentData` which allowed me to refactor
```
export function flattenPunishment(punishment: PunishmentDataWithSchedule): PunishmentData
```
to
```
export function flattenPunishment(punishment: PunishmentDataWithSchedule): PunishmentDataV2
```
The API isn't expecting it (see [PunishmentsController.PunishmentRequest](https://github.com/ministryofjustice/hmpps-manage-adjudications-api/blob/3e95d136b89f1ef586a6b5b32f40990c3b33f9a5/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsmanageadjudicationsapi/controllers/reported/PunishmentsController.kt#L40)) - so think we're ok here?

neilmendum: Added that back in with https://github.com/ministryofjustice/hmpps-manage-adjudications/pull/760/commits/dac99d5d70524114bbbca096f1ade80a09a49865

`consecutiveReportAvailable` is used in a couple of places - we can look to refactor the use when we tackle NN-5370

timoj58 (approved):

server/data/PunishmentResult.ts
neilmendum: Fixed with https://github.com/ministryofjustice/hmpps-manage-adjudications/pull/761

PR #764, opened 2023-08-07
NN-5174: remove DATA_INSIGHTS_FLAG and caseload check

neilmendum:
https://dsdmoj.atlassian.net/browse/NN-5174

### DIFF
--- a/integration_tests/integration/home.cy.ts
+++ b/integration_tests/integration/home.cy.ts
@@ -2,7 +2,6 @@ import moment from 'moment'
 import adjudicationUrls from '../../server/utils/urlGenerator'
 import HomepagePage from '../pages/home'
 import Page from '../pages/page'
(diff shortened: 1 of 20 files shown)


integration_tests/integration/home.cy.ts
```diff
@@ -149,33 +148,7 @@ context('Home page', () => {
...
-    ] as CaseLoad[])
-    cy.task('stubAuthUser', {
-      username: 'USER1',
-      activeCaseLoadId: 'RNI',
-    })
-    cy.signIn()
-  })
-
-  it('should link to the correct location - adjudication data', () => {
+  it('should link to the correct location - data insights', () => {
```
neilmendum, line 151: As we're now enabling it for all caseloads, moved this test to be within `context('Home page'`

timoj58 (approved):

PR #766, opened 2023-08-08
NN-5419: split adjudications API client into system and user token instances

neilmendum:
https://dsdmoj.atlassian.net/browse/NN-5419


server/services/hearingsService.ts
```diff
@@ -49,18 +50,24 @@ export default class HearingsService {
       reason: adjournReason,
       plea,
     }
-    return new ManageAdjudicationsClient(user).createAdjourn(chargeNumber, hearingOutcomeDetails)
+    return new ManageAdjudicationsUserTokensClient(user).createAdjourn(chargeNumber, hearingOutcomeDetails)
   }
 
   async getHearingOutcome(chargeNumber: string, user: User) {
-    const adjudication = await new ManageAdjudicationsClient(user).getReportedAdjudication(chargeNumber)
+    const token = await this.hmppsAuthClient.getSystemClientToken(user.username)
```
neilmendum, line 57: To construct this `token` we make call to HMPPS Auth and pass the [clientToken](https://github.com/ministryofjustice/hmpps-manage-adjudications/blob/9a701b305c0576870bcaefc485942b7760c8a33c/server/data/hmppsAuthClient.ts#L13-L16) which corresponds to the `manage-adjudications-client` client credentials

server/services/reportedAdjudicationsService.ts
```diff
@@ -328,7 +343,8 @@ export default class ReportedAdjudicationsService {
     details: string,
     user: User
   ): Promise<ReportedAdjudicationResult> {
-    return new ManageAdjudicationsClient(user).updateAdjudicationStatus(chargeNumber, {
+    const token = await this.hmppsAuthClient.getSystemClientToken(user.username)
+    return new ManageAdjudicationsSystemTokensClient(token, user).updateAdjudicationStatus(chargeNumber, {
```
timoj58, line 347: i think this should be user, as its the ALO review endpoint

timoj58 (approved):

PR #769, opened 2023-08-10
NN-5461

natclamp-moj:


integration_tests/pages/punishment.ts
```diff
@@ -2,7 +2,7 @@ import Page, { PageElement } from './page'
 
 export default class PunishmentPage extends Page {
   constructor() {
-    super('Add a new punishment')
+    super('Add a')
```
natclamp-moj, line 5: The title is now different in different circumstances so I've just included the common part!

timoj58 (approved):

PR #804, opened 2023-09-01
NN-5529 Modified Datasource and name for prc

nagendramoj:


server/views/pages/dataInsights/protectedAndResponsivityCharacteristicsTab.njk
```diff
@@ -11,7 +11,7 @@
 {% set title = "Adjudication data - Protected and responsivity characteristics" %}
 
 {% block pageTitle %}
-  {{ title }} - Adjudications - Digital Prison Services
```
neilmendum, line 14: Don't think we want to update this - it would revert https://github.com/ministryofjustice/hmpps-manage-adjudications/pull/789

neilmendum (approved):
LGTM

PR #807, opened 2023-09-05
NN-5381: display DIS 1/2 form issue history

neilmendum:
https://dsdmoj.atlassian.net/browse/NN-5381

### DIFF
--- a/server/views/pages/adjudicationForReport/formsTab.njk
+++ b/server/views/pages/adjudicationForReport/formsTab.njk
@@ -39,6 +39,13 @@
 
     {% set rows = [] %}
     {% for report in reports %}
+      {% for disIssue in report.formattedDisIssueHistory %}
+        {% set rows = (rows.push([
+          { text: disIssue.formattedDateTimeOfIssue or ' - ' },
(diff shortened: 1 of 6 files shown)


server/views/pages/adjudicationForReport/formsTab.njk
```diff
@@ -39,6 +39,13 @@
 
     {% set rows = [] %}
     {% for report in reports %}
+      {% for disIssue in report.formattedDisIssueHistory %}
```
neilmendum, line 42: In the designs the table is displayed ascending hence me putting the history (if there is any) before the latest issuance. Also checked in the API and [ReportedAdjudicationBaseService returns](https://github.com/ministryofjustice/hmpps-manage-adjudications-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsmanageadjudicationsapi/services/reported/ReportedAdjudicationBaseService.kt#L252-L258) `DisIssueHistory` in ascending order so we can simply display it in the order we get it 👌

natclamp-moj (approved):

PR #813, opened 2023-09-07
NN-5540 NN-5541: creating reports on behalf of someone else - user entry and submit

neilmendum:
https://dsdmoj.atlassian.net/browse/NN-5540
https://dsdmoj.atlassian.net/browse/NN-5541

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -20,7 +20,6 @@ export default defineConfig({
   reporterOptions: {
     configFile: 'reporter-config.json',
   },
-  videoUploadOnPasses: false,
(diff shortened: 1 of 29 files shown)


cypress.config.ts
```diff
@@ -20,7 +20,6 @@ export default defineConfig({
   reporterOptions: {
     configFile: 'reporter-config.json',
   },
-  videoUploadOnPasses: false,
```
neilmendum, line 23: Removed in Cypress 13 - see https://github.com/cypress-io/cypress/issues/26899

timoj58 (approved):

PR #826, opened 2023-09-20
NN-5558 NN-5566 NN-5567 NN-5568

natclamp-moj:
NN-5558 NN-5566 NN-5567 NN-5568

### DIFF
--- a/server/views/macros/activateSuspendedPunishmentsTable.njk
+++ b/server/views/macros/activateSuspendedPunishmentsTable.njk
@@ -16,7 +16,7 @@
             classes: "columnWidth30 verticalAlign"
           },
           {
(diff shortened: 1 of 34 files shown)


server/views/macros/activateSuspendedPunishmentsTable.njk
```diff
@@ -16,7 +16,7 @@
             classes: "columnWidth30 verticalAlign"
           },
           {
-            text: punishment.reportNumber,
+            text: punishment.chargeNumber,
```
natclamp-moj, line 19: this is actually a bug which must be in live from when we switched from report number to charge number!

neilmendum (approved):
LGTM

PR #854, opened 2023-10-04
NN-5502: create 2 versions of DIS 1/2: prisoner and internal

neilmendum:
https://dsdmoj.atlassian.net/browse/NN-5502


server/routes/disPdf/dis12Pdf.ts
```diff
@@ -41,7 +57,7 @@ export default class Dis12Pdf {
       `pages/noticeOfBeingPlacedOnReportFooter`,
       {},
       {
-        filename: `adjudication-report-${chargeNumber}`,
+        filename: `notice-of-being-placed-on-report-${chargeNumber}.pdf`,
```
neilmendum, line 60: Updated this to be consistent with other DIS forms we've added since DIS 1/2 was originally built

natclamp-moj (approved):

PR #1056, opened 2024-04-11
NN-5941

natclamp-moj:


server/views/pages/offenceCodeProtectedCharacteristics.njk
```diff
@@ -0,0 +1,106 @@
...
+    },
+    {
+      value: ProtectedCharacteristicsTypes.SEX,
+      text: "Sex"
+    },
+    {
+      value: ProtectedCharacteristicsTypes.SEX_ORIENTATION,
+      text: "Sexual orientation"
+    }
+  ] %}
```
natclamp-moj, line 60: Could set these in the controller and iterate through in the njk if that's preferable

server/views/pages/offenceCodeProtectedCharacteristics.njk
```diff
@@ -0,0 +1,106 @@
...
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+  <form method="POST" novalidate="novalidate" class="govuk-!-margin-top-5" data-qa="offence-code-protected-characteristics-form">
+    <input type="hidden" name="_csrf" value="{{ csrfToken }}"/>
+    <input type="hidden" name="protectedCharacteristicPage" value="true"/>
```
natclamp-moj, line 66: This is a flag so we can tell which njk we're rendering if there's a validation error so we know what to check for

server/views/pages/offenceCodeProtectedCharacteristics.njk
natclamp-moj: Might be able to jsut send through the enum and then use a njk filter to set the text based off the enum

timoj58 (approved):

PR #1192, opened 2024-11-17
MAP-1808 use locationsApi instead of prisonApi to get prison internal …

GurnankCheema:
- this code change is to use locationsApi instead of prisonApi to get prison internal locations particularly for components like 'select drop-downs' but also when persisted data comes from the backend, its locationId is converted and then locationsApi is called to get the full location details

- locationId's persisted (via the hmpps-manage-adjudications-api) are still in the prisonApi format, i.e the integer form of locationId, so backend services will not be affected. 

- the nomisSyncPrisonerMappingApi is used to obtain prisonApi locationId using the locationsInsidePrisonApi locationId and vice-versa

- here are the prisonApi endpoints that are now being provided by locationsInsidePrisonApi
    - getLocation
    - getLocations
    - getAdjudicationLocations
       
- the locationsInsidePrisonApi returns locationIds in UUID form. The data object returned is also different to that returned by prisonApi and the two map in the following way
   - locationId            -> id 
   - agencyId             -> prisonId
   - locationPrefix     -> key
   - userDescription -> localName
  
- there is still work to be done to fine tune the response from the locationsInsidePrisonApi, specifically the casing of location descriptions and also the filtering of the locations returned. This PR is marked as 'draft' to prevent an inadvertent release before the locationsApi has been fixed.

### DIFF
--- a/integration_tests/integration/hearingTab.cy.ts
+++ b/integration_tests/integration/hearingTab.cy.ts
@@ -2119,186 +2119,186 @@ context('Hearing details page', () => {
       hearingTabPage.removeOutcomeButton().should('not.exist')
     })
   })
-  // describe('Test scenarios - reporter view', () => {
-  //   ;[
-  //     { id: '1524480' },
-  //     { id: '1524493' },
-  //     { id: '1524494' },
-  //     { id: '1524495' },
-  //     { id: '1524496' },
-  //     { id: '1524497' },
-  //     { id: '1524502' },
-  //     { id: '1524504' },
-  //   ].forEach(adj => {
-  //     it('should contain the required page elements', () => {
-  //       cy.visit(adjudicationUrls.hearingDetails.urls.report(adj.id))
-  //       const hearingTabPage = Page.verifyOnPage(hearingTab)
-  //       hearingTabPage.reviewStatus().should('exist')
-  //       hearingTabPage.removeHearingButton().should('not.exist')
-  //       hearingTabPage.viewAllCompletedReportsLink().should('not.exist')
-  //       hearingTabPage.ReturnToAllHearingsLink().should('not.exist')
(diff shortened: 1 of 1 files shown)


server/routes/adjudicationForReport/scheduleHearing/scheduleHearingValidation.ts
```diff
@@ -3,7 +3,8 @@ import { formatDate, formatTimestampToDate, formatTimestampToTime } from '../../
 
 type ScheduleHearingForm = {
   hearingDate?: SubmittedDateTime
-  locationId?: number
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  locationId?: any
```
dafhoustonmoj, line 7: is any correct here?

integration_tests/integration/printDis12Report.cy.ts
```diff
@@ -64,14 +64,24 @@ context('Print notice of being placed on report (DIS 1 and 2)', () => {
         },
       ],
     })
-    cy.task('stubGetLocation', {
-      locationId: 25538,
-      response: testData.residentialLocations()[0],
-    })
+    // cy.task('stubGetLocation', {
```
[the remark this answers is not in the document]
GurnankCheema: yes, I forgot to remove it.

server/routes/adjudicationForReport/scheduleHearing/scheduleHearingValidation.ts
GurnankCheema: My mistake it should be string only. 
The ScheduleHearingForm is used in the validation of `/schedule-hearing/:chargeNumber` when posting. 
The select drop-down on this page now gets its entries from the locations-api which will have UUIDs for the locationId

dafhoustonmoj (approved):
LGTM

PR #1202, opened 2024-12-04
MAP-1932 use correct endpoint to get nomis mapping data

GurnankCheema:
previously used incorrect endpoint. Both old and new return same data.
New endpoint requires NOMIS_DPS_MAPPING__LOCATIONS__R role which is being requested. 

will not merge until new role in place for all 3 environments


Mjwillis (approved):
LGTM, you'll need the role adding

PR #1206, opened 2024-12-13
Updated redis from 3.1.2 to 4

cugorji-justice:
https://dsdmoj.atlassian.net/browse/NN-5989


server/data/tokenStore.ts
```diff
@@ -1,21 +1,18 @@
...
   public async setToken(key: string, token: string, durationSeconds: number): Promise<void> {
-    this.setRedisAsync(key, token, 'EX', durationSeconds)
+    // For redis v4, the expiration is set via an options object:
+    await (await this.client).set(key, token, { EX: durationSeconds })
   }
 
-  public async getToken(key: string): Promise<string> {
-    return this.getRedisAsync(key)
+  public async getToken(key: string): Promise<string | null> {
+    return (await this.client).get(key)
```
andrewrlee, line 16: think there shouldn't be an await on the client its self

`await this.client.get(key)`
`await this.client.set(key, token, { EX: durationSeconds })

server/data/redisClient.ts
```diff
@@ -5,16 +5,57 @@ import config from '../config'
...
+        logger.info(`Retry Redis connection attempt: ${attempts}, next attempt in: ${nextDelay}ms`)
+        return nextDelay
+      },
+    },
   })
 
   client.on('error', (e: Error) => logger.error('Redis client error', e))
 
-  return client
+  client.connect().catch((e: Error) => logger.error('Failed to connect to Redis', e))
```
andrewrlee, line 58: maybe do this in the setUpWebSession [like in the template project](https://github.com/ministryofjustice/hmpps-template-typescript/blob/576409caf776c876219b897c715dc329ca30add4/server/middleware/setUpWebSession.ts#L13)

.circleci/config.yml
```diff
@@ -121,6 +121,7 @@ jobs:
       - run:
           name: Wait for node app to start
           command: sleep 5
+#          command: npx wait-on http://localhost:3007
```
andrewrlee, line 124: You can use the command from the helm orb https://github.com/ministryofjustice/offender-categorisation/blob/475a15f6e9ad0c237c8ad413fd8b73c8a913ffd9/.circleci/config.yml#L127-L129

server/data/tokenStore.ts
```diff
@@ -1,21 +1,20 @@
...
-    this.getRedisAsync = promisify(redisClient.get).bind(redisClient)
-    this.setRedisAsync = promisify(redisClient.set).bind(redisClient)
+  constructor(redisClient = createRedisClient()) {
+    this.client = redisClient
   }
 
   public async setToken(key: string, token: string, durationSeconds: number): Promise<void> {
-    this.setRedisAsync(key, token, 'EX', durationSeconds)
+    // For redis v4, the expiration is set via an options object:
+    this.client.set(this.prefix + key, token, { EX: durationSeconds })
```
andrewrlee, line 14: ooh, if you don't want to await it then you'd need to return this promise.

.circleci/config.yml
```diff
@@ -118,9 +118,13 @@ jobs:
           name: Run the node app.
           command: npm run start-feature
           background: true
-      - run:
-          name: Wait for node app to start
-          command: sleep 5
+#      - run:
```
theDustRoom, line 121: Can you remove the commented code as it's now redundant?

theDustRoom (changes requested):
Just a minor change, all looks really good.

theDustRoom (approved):
Nice work Chux

PR #1219, opened 2025-01-10
added log to see the data returned

cugorji-justice:
https://dsdmoj.atlassian.net/browse/NN-6020


theDustRoom (approved):
LGTM - These will have to be removed before we move into PreProd and Prod

PR #1243, opened 2025-02-24
Move to new alerts data API endpoint

cugorji-justice:
https://dsdmoj.atlassian.net/browse/NN-6029


natclamp-moj:
I think you need to add the new API url to the feature.env file
You can also [remove the API call](https://github.com/ministryofjustice/hmpps-manage-adjudications/blob/35f0e2fe292c251a69d6a1faa37bd0c2ebb52d15/server/data/prisonApiClient.ts#L124) to get alerts from prisonAPI, I think

theDustRoom (approved):
LGTM

PR #1256, opened 2025-03-20
NN-6041 - Global Search

theDustRoom:
Enforce GlobalSearch rule
Patched route `/search-for-prisoner?transfer=true`
Patched route `/select-prisoner?searchTerm=SOMETHING&transfer=true`
Refactored and simplified code
DRY fixes


integration_tests/integration/notificationOfBeingPlacedOnReport.cy.ts
theDustRoom, line 1: This test is flaky. Temporarily disabled this test and will raise a ticket to look into this later.

PR #1264, opened 2025-04-08
MAP-2119: Adding UUID parameter for calls to adjudications API

ashleygyngell:
Adding UUID parameter for all calls to adjudications API. Both` locationId `and `locationUUID` will be sent for the time being, with a plan to remove` locationId` once some other work has been completed down the line.


cugorji-justice (approved):
LGTM

PR #1267, opened 2025-04-10
NN-6061 Add information banner to Adjudications data pages

eve-bramley:
added information banners to Adjudications data pages


cugorji-justice (approved):
LGTM

PR #1268, opened 2025-04-16
NN-6061 Add information banner to Adjudications data pages

eve-bramley:
Updated alert from html to nunjucks


server/views/pages/dataInsights/offenceTypeTab.njk
```diff
@@ -16,6 +17,15 @@
   {{ giveFeedbackBanner() }}
 
   {{ breadcrumb() }}
+
```
theDustRoom, line 20: Can you indent this block so it's the same as the `{{ breadcrumb() }}` above?

theDustRoom (dismissed):
Nice one, thanks Eve

theDustRoom (approved):

PR #1386, opened 2025-12-01
LSM-192 - PATCH - Allow Cypress to install binary

theDustRoom:
Cypress needs to install the binary, this is disabled in the Dockerfile though


sophiemetcalf (approved):
LGTM!

PR #1419, opened 2026-03-20
Backout change to Dockerfile

theDustRoom:
The base image looks to have been patched so this change is no longer required


eve-bramley (approved):
LGTM

PR #1423, opened 2026-04-28
LSM-445 - Maintenance Mode fix

theDustRoom:
Fix active caseload fallback and maintenance redirect

Use the session user’s active caseload when frontend component metadata is unavailable, preventing downstream API calls from missing the Active-Caseload header.

Replace the maintenance mode wildcard route with pathless middleware so the redirect works with Express 5 routing.


eve-bramley (approved):
LGTM

PR #1444, opened 2026-06-09
Run integration tests in parallel using shared workflow

theDustRoom:
Use the parallel integration-test workflow from hmpps-github-actions.

Run Cypress integration tests across six shards and replace the deprecated Redis-specific integration-test workflow.


Mjwillis (approved):
LGTM

PR #1445, opened 2026-06-10
IR-1675 - Upgrade to TypeScript 6

theDustRoom:
Upgrade TypeScript 6 configuration and fix compatibility issues
Align the TypeScript configuration with the HMPPS template, typecheck integration tests, and update code and test fixtures for NodeNext and TypeScript 6 compatibility.


integration_tests/tsconfig.json
```diff
@@ -1,11 +1,13 @@
...
   "compilerOptions": {
-    "target": "es5",
-    "ignoreDeprecations": "6.0",
     "noEmit": true,
     "lib": ["es5", "dom", "es2015.promise"],
+    "noImplicitAny": false,
+    "typeRoots": ["../server/@types", "../node_modules/@types", "../node_modules"],
     "types": ["cypress"],
-    "esModuleInterop": true
+    "skipLibCheck": true
```
xoen: now that you're extending from the top-level `tsconfig.json` this line can probably go as it's the same as in the parent config file.

tsconfig.json
```diff
@@ -1,25 +1,21 @@
...
     "resolveJsonModule": true,
-    "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
     "noImplicitAny": true,
-    "strictNullChecks": false,
-    "strictFunctionTypes": false,
-    "useUnknownInCatchVariables": false,
-    "experimentalDecorators": true
+    "isolatedModules": true,
+    "experimentalDecorators": true,
```
xoen, line 15: I don't know if it applies to this project, but what I've found in the projects I upgraded to TS6 is that this `experimentalDecorators` options was set to true but it wasn't really used anywhere.

You may find that type checking and everything still works when removing this option (unless you genuinely use these in this project).

(from the little I know, these are something that was added in the past but these days one should favour standard JS decorators instead, if using any)

xoen (dismissed):
FWIW LGTM.

tsconfig.json
theDustRoom: Yes, it is still required. The application uses class-transformer’s legacy @Expose() decorators in the prisoner result models. Removing experimentalDecorators causes TypeScript 6 decorator-signature errors and would require migrating or replacing those transformations. I’ve therefore retained the existing setting as part of this upgrade.

xoen: Ah ok, in that case you need that option :) (and I agree migrating would be overkill as part of this)

integration_tests/tsconfig.json
theDustRoom: Good catch 👍

xoen (approved):

PR #1448, opened 2026-06-17
IR-1692: Upgrade govuk and moj frontend

theDustRoom:
Upgrade govuk-frontend and `@ministryofjustice/frontend`, migrating Sass imports to `@use` and aligning asset loading with the newer frontend packages.

Remove jQuery and the unused legacy jQuery UI datepicker path, while keeping the active HMPPS datepicker component in place.

Update affected frontend markup to use GOVUK 6 compatible classes explicitly, including links, compact labels, lists and the homepage layout wrapper where GOVUK no longer applies the same implicit global styles.

Preserve the existing homepage card behaviour and prisoner playback styling after the GOVUK template/layout changes.

Tidy related CSP/static asset config and retain only the npm overrides needed to keep the audit clean.


cugorji-justice (approved):
LGTM

PR #1449, opened 2026-06-23
Removed duplicated request

xoen:
The `reviewerEditOffenceWarning` request handler was making the same request to get the information about the Adjudication twice:
- [in the request handler](https://github.com/ministryofjustice/hmpps-manage-adjudications/blob/2c3036e1d36c672d34792c9fdf202a2da4e7d995/server/routes/reviewerEditOffenceWarning/reviewerEditOffenceWarning.ts#L26) itself
- [in the `DecisionTreeService`](https://github.com/ministryofjustice/hmpps-manage-adjudications/blob/2c3036e1d36c672d34792c9fdf202a2da4e7d995/server/services/decisionTreeService.ts#L61-L63)

Assuming this `GET` request has no side effects we can avoid the duplicated request, as the adjudication data is returned by the `reportedAdjudicationIncidentData()` method anyway (it's probably not an heavy HTTP request but every little helps)

Also:
- renamed `reportedAdjudicationService` (singular) instance property to `reportedAdjudicationsService` (plural) for consistency, every other place uses the plural form and it was surprising.
- tweaked `offenceQAndASummaryList()` Nunjucks macro to remove unnecessary `value` variable: It reads better without the layer of indirection and it makes the relation between key/question and value/answer more obvious
- changed type imports in `reviewerEditOffenceWarning` request handler to be type-only imports

### DIFF
--- a/server/routes/reviewerEditOffenceWarning/reviewerEditOffenceWarning.ts
+++ b/server/routes/reviewerEditOffenceWarning/reviewerEditOffenceWarning.ts
@@ -1,9 +1,9 @@
-import { Request, Response } from 'express'
-import DecisionTreeService from '../../services/decisionTreeService'
-import UserService from '../../services/userService'
+import type { Request, Response } from 'express'
+import type DecisionTreeService from '../../services/decisionTreeService'
+import type UserService from '../../services/userService'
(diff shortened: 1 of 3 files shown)


theDustRoom (approved):
Nice catch Aldo

PR #1452, opened 2026-06-29
IR-1715 Prevent selecting a consecutive punishment that creates a loop

Mjwillis:
## What

Stop users from making two charges mutually consecutive (a \"consecutive loop\") on the *\"Which punishment will it be consecutive to?\"* page.

## Why

Looped consecutive punishments break downstream services such as Calculate Release Dates. The API now rejects them with a 400, but the frontend only POSTs punishments on the final \"check punishments\" page — so without a frontend guard the user wouldn't see the error until the very end. This change prevents the user reaching that state.

## How

- Move the loop-prevention filter into \`getPossibleConsecutivePunishments()\` in \`PunishmentsService\` so all callers automatically receive a loop-safe list (addressing review feedback: no future caller can accidentally re-introduce the bug or forget to filter).
- \`validateForm\` is updated to check presence in the pre-filtered list rather than inspecting \`consecutiveChargeNumber\` on individual entries, keeping the defence-in-depth POST guard intact.
- Added \`whichPunishmentConsecutiveToValidation.ts\` plus a submit-time guard that re-renders the page with a \`govukErrorSummary\` if a looping selection is somehow forced.
- Added the error summary block to the page template.

## Tests

- Unit test in \`punishmentsService.test.ts\` covering the loop filter directly in the service.
- Route tests covering the GET filter (looping charge hidden) and the POST guard (selection rejected, nothing saved to session).
- Unit tests for the validation module.

### DIFF
--- a/server/routes/punishmentsForReport/additionalDays/whichPunishmentConsecutiveTo/whichPunishmentConsecutiveToPage.ts
+++ b/server/routes/punishmentsForReport/additionalDays/whichPunishmentConsecutiveTo/whichPunishmentConsecutiveToPage.ts
@@ -5,6 +5,9 @@ import { hasAnyRole } from '../../../../utils/utils'
 import adjudicationUrls from '../../../../utils/urlGenerator'
 import PunishmentsService from '../../../../services/punishmentsService'
 import { PrivilegeType, PunishmentType, RehabilitativeActivity } from '../../../../data/PunishmentResult'
+import { FormError } from '../../../../@types/template'
+import { ConsecutiveAdditionalDaysReport } from '../../../../data/manageAdjudicationsUserTokensClient'
+import validateForm from './whichPunishmentConsecutiveToValidation'
 
 export enum PageRequestType {
   CREATION,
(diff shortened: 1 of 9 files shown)


server/routes/punishmentsForReport/additionalDays/whichPunishmentConsecutiveTo/whichPunishmentConsecutiveToPage.ts
```diff
@@ -49,18 +58,34 @@ export default class WhichPunishmentConsecutiveToPage {
 
     return res.render(`pages/whichPunishmentConsecutiveTo.njk`, {
       cancelHref: adjudicationUrls.awardPunishments.urls.modified(chargeNumber),
-      possibleConsecutivePunishments,
+      // hide any charge already consecutive to this one - selecting it would create a loop
+      possibleConsecutivePunishments: possibleConsecutivePunishments.filter(
+        report => report.punishment.consecutiveChargeNumber !== chargeNumber,
```
xoen: Is there any scenario where we'd need the punishments that would create a loop?

In other words, I'm wondering whether this filtering-out should happen in the service inside `getPossibleConsecutivePunishments()` instead? 🤔 

This would prevent accidentally re-introduce the loop in some other place using that method (or having possible loops in places that forget to filter these out).

Mjwillis: I'll take a look at where else its used

Mjwillis: Its not used anywhere else, I can make the changes if you want... but it doesn't give you much

Mjwillis: @xoen updated

Mjwillis: Just fixing a test...

xoen (approved):
FWIW LGTM

PR #1453, opened 2026-07-01
Adds missing "govuk-body" classes

simonnebMOJ:
Noticed a few areas in the service where paragraph text wasnt rendering correctly. This PR should fix those by adding the `govuk-body` to the classes.


jonesj2193 (approved):
LGTM

PR #1457, opened 2026-07-10
IR-1758/IR-1769: Increased max adults' added days/prospective added days to 84

xoen:
To reflect policy changes.

NOTE: The corresponding maximum number of days for YOI is unchanged. Also, we didn't change the forms that mention these numbers.


cugorji-justice (approved):
LGTM

PR #1461, opened 2026-07-16
IR-1745/DIS1-2-policy-updates

ievamaz:
amended DIS1/2 form with policy updates
fixed h3 headers and moved adjournments to start of the form


eve-bramley (approved):
LGTM!

PR #1463, opened 2026-07-16
IR-1753: Increase loss of privileges max number of days for adults (to 84)

xoen:
The `PunishmentType.PRIVILEGE` type is for loss of privileges like Canteen, TV, Other, etc...

**NOTE**: This is not changing the validation on `PunishmentType.EXTRA_WORK` or `PunishmentType.REMOVAL_ACTIVITY`, I asked Steve for clarification but not touching those as part of this PR. Also, the Stoppage of earnings punishment limit was already 84 days for adults so no change there neither.


eve-bramley (approved):
LGTM

PR #1464, opened 2026-07-17
IR-1780 Allow deploying a feature branch to the dev environment

Mjwillis:
## Why

We have a long-running feature branch (~2 months) that can't go live yet but needs to be deployed to **dev** for testing. The frontend pipeline previously only published a Docker image on `main` and had no manual deploy workflow, so there was no way to get a feature-branch build onto dev. (The API repo already supports this.)

## What

**`pipeline.yml`**
- `check_publish.enabled` now also turns on when the pipeline is run via **Run workflow** (`workflow_dispatch`) with `push=true` on any branch, so a feature-branch build publishes a signed, versioned image to `ghcr.io`. Normal `main` and normal (non-dispatch) feature-branch behaviour are unchanged.
- `deploy_dev` / `deploy_preprod` / `deploy_prod` are now gated on `if: github.ref == 'refs/heads/main'` instead of the publish flag. Publishing an image no longer implies a deploy — a feature-branch dispatch can never auto-deploy anywhere.

**`deploy_to_env.yml`** (new)
- Manual `workflow_dispatch` deploy mirroring the API's: pick `environment` (dev/preprod/prod) and a `version`, runs helm lint, then the shared `deploy_env.yml@v2`.

## How to use

1. Actions → **Pipeline** → *Run workflow* → select the feature branch → set `push = true`. Note the version from the `build` job's `app_version` output.
2. Actions → **Deploy to environment** → *Run workflow* → `environment = dev`, `version = <that version>`.

Note: deploying a feature branch to dev replaces what's running there; re-deploy the latest `main` version to dev when finished testing.

### DIFF
--- a/.github/workflows/deploy_to_env.yml
+++ b/.github/workflows/deploy_to_env.yml
@@ -1,0 +1,40 @@
+name: Deploy to environment
+
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: Environment
+        type: choice
+        required: true
+        options:
+          - dev
+          - preprod
+          - prod
+        default: 'dev'
+      version:
+        description: version to be deployed to the environment - must already exist.
+        required: true
+        default: ''
+        type: string
+
+permissions:
+  contents: read
+
+jobs:
+  helm_lint:
+    name: helm lint
+    uses: ministryofjustice/hmpps-github-actions/.github/workflows/test_helm_lint.yml@v2 # WORKFLOW_VERSION
+    secrets: inherit
+    with:
+      environment: ${{ inputs.environment }}
+  deploy_env:
+    name: Deploy to environment
+    needs:
+      - helm_lint
+    uses: ministryofjustice/hmpps-github-actions/.github/workflows/deploy_env.yml@v2 # WORKFLOW_VERSION
+    secrets: inherit
+    with:
+      environment: ${{ inputs.environment }}
+      app_version: ${{ inputs.version }}
+      slack_notification: 'false' # if 'true', it needs a valid (NON)PROD_RELEASES_SLACK_CHANNEL variable to be set
--- a/.github/workflows/deploy_to_env.yml
+++ b/.github/workflows/deploy_to_env.yml
@@ -10,7 +10,8 @@ on:
         options:
           - dev
           - preprod
-          - prod
+          # Deploy to Production only from `main` branch, using normal process
+          # - prod
         default: 'dev'
       version:
         description: version to be deployed to the environment - must already exist.
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -34,10 +34,10 @@ jobs:
     name: Check publish flag
     runs-on: ubuntu-latest
     outputs:
-      enabled: ${{ github.ref == 'refs/heads/main' && (inputs.push || inputs.push == null) }}
+      enabled: ${{ (github.ref == 'refs/heads/main' && (inputs.push || inputs.push == null)) || (github.event_name == 'workflow_dispatch' && inputs.push == true) }}
     steps:
       - name: Set publish flag
-        run: echo "Publishing enabled ${{ github.ref == 'refs/heads/main' && (inputs.push || inputs.push == null) }}"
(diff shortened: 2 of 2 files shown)


.github/workflows/pipeline.yml
```diff
@@ -34,10 +34,10 @@ jobs:
     name: Check publish flag
```
xoen, line 34: What are the changes in this file?

I see this would deviate from what we do in something like [Incident Reporting](https://github.com/ministryofjustice/hmpps-incident-reporting/blob/c7a049ba94a742e2cfe8ef5e5ca47c679c38e492/.github/workflows/pipeline.yml#L41) or the [TS template](https://github.com/ministryofjustice/hmpps-template-typescript/blob/c0d4c2607e657932e89a91f972354fa0b39b99dd/.github/workflows/pipeline.yml) - which I'd rather avoid unless there is a truly strong argument for it.

Does this file need to change to add the `deploy_to_env` workflow? (which is word for word like the one in IRS or TS template, which is good)

.github/workflows/pipeline.yml
```diff
@@ -34,10 +34,10 @@ jobs:
...
     runs-on: ubuntu-latest
     outputs:
-      enabled: ${{ github.ref == 'refs/heads/main' && (inputs.push || inputs.push == null) }}
+      enabled: ${{ (github.ref == 'refs/heads/main' && (inputs.push || inputs.push == null)) || (github.event_name == 'workflow_dispatch' && inputs.push == true) }}
```
xoen, line 37: So, I understand this correctly this is so that we can manually trigger the run of this workflow and pass `push = true` which is what determine whether the built docker image will be pushed to the Docker registry (GHCR).

Currently it only pushes if the branch is main. While after this change it would use the provided value for `push` if the workflow was triggered manually (`workflow_dispatch`).

THEN, after the image is pushed to GHCR, we can run the `deploy_to_env` manually to deploy the given image tag (pushed from step above) to `dev`.

.github/workflows/pipeline.yml
```diff
@@ -81,7 +81,7 @@ jobs:
       sign_image: ${{ needs.check_publish.outputs.enabled == 'true' }}
   deploy_dev:
     name: Deploy to the dev environment
-    if: needs.check_publish.outputs.enabled == 'true'
+    if: github.ref == 'refs/heads/main'
```
xoen, line 84: So these changes are so that triggering the pipeline manually with `push == true` doesn't also deploy to `dev` 🤔 (or `preprod` below).

Effectively it's changing the meaning of push from "push to GHCR _**and**_ to k8s" to "just push to GHCR"?

Is it mainly a safety guard to prevent accidental deployment of a branch to `dev` or possibly other environments? I guess it would be convenient to deploy to `dev` in one go by manually running this workflow, although I do appreciate having that guard for `preprod`/`prod` 🤔

xoen: On the flip side - _**if**_ I'm reading this correctly - leaving the `if: needs.check_publish.outputs.enabled == 'true'` **for `dev`** would mean we don't need to separately run the `deploy_to_env` workflow to deploy a branch to dev, and we could just run the pipeline for the given branch with `push == true` and that would also deploy to `dev`?

.github/workflows/deploy_to_env.yml
```diff
@@ -0,0 +1,40 @@
...
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: Environment
+        type: choice
+        required: true
+        options:
+          - dev
+          - preprod
+          - prod
```
theDustRoom: Should prod be available in this workflow?

My understanding is that branch deployments should only be able to target dev and preprod, with prod releases only happening from main via the normal release flow/environment approvals. If that’s right, I think we should remove prod from this dropdown so non-main branches don’t even have the option.

theDustRoom (approved):
LGTM

.github/workflows/deploy_to_env.yml
xoen: Sorry, completely missed this comment somehow.

I made this change in [this commit]([Prevent manual deployment to production](https://github.com/ministryofjustice/hmpps-manage-adjudications/pull/1464/commits/d118154fbc12e01fa15c16b2daac1024f81b469a)), resolving this. Thanks for the review!

PR #1466, opened 2026-07-21
IR 1757 recording new visits punishment

cugorji-justice:
### Summary

Adds support for the new Adult Rule 51 social-visits punishments:

- Adds `Loss of social visits` and `Restriction of social visits` as separate punishment options
- Makes both options available for adult adjudications only
- Requires the user to record whether the prisoner has children under 18
- Preserves the Yes/No answer throughout create, edit, suspension and activation journeys
- Enforces the policy limits:
  - Loss of social visits: 27 days
  - Restriction of social visits: 84 days
- Displays the required validation messages when these limits are exceeded
- Keeps the invalid duration available for correction
- Calculates and displays the inclusive Last Day from the selected start date and duration
- Displays the child-under-18 answer in punishment summaries and tables
- Preserves existing behaviour for non-visits punishments

### DIFF
--- a/server/data/PunishmentResult.ts
+++ b/server/data/PunishmentResult.ts
@@ -11,6 +11,8 @@ export enum PunishmentType {
   CAUTION = 'CAUTION',
   DAMAGES_OWED = 'DAMAGES_OWED',
   PAYBACK = 'PAYBACK',
+  RESTRICTION_OF_SOCIAL_VISITS = 'RESTRICTION_OF_SOCIAL_VISITS',
+  LOSS_OF_SOCIAL_VISITS = 'LOSS_OF_SOCIAL_VISITS',
 }
 
 export enum PrivilegeType {
@@ -65,6 +67,7 @@ export type PunishmentData = {
   consecutiveChargeNumber?: string
   consecutiveReportAvailable?: boolean
   damagesOwedAmount?: number
+  hasChildUnder18?: boolean
   canRemove?: boolean
   canEdit?: boolean
   rehabilitativeActivities: RehabilitativeActivity[]
@@ -95,6 +98,7 @@ export type PunishmentDataWithSchedule = {
   consecutiveChargeNumber?: string
   consecutiveReportAvailable?: boolean
   damagesOwedAmount?: number
+  hasChildUnder18?: boolean
   canRemove?: boolean
   canEdit?: boolean
   rehabilitativeActivities: RehabilitativeActivity[]
@@ -279,6 +283,10 @@ export function convertPunishmentType(
       return 'Recovery of money for damages'
     case PunishmentType.PAYBACK:
       return 'Payback punishment'
+    case PunishmentType.RESTRICTION_OF_SOCIAL_VISITS:
+      return 'Restriction of social visits'
+    case PunishmentType.LOSS_OF_SOCIAL_VISITS:
+      return 'Loss of social visits'
     default:
       return null
   }
@@ -297,6 +305,7 @@ export function flattenPunishment(punishment: PunishmentDataWithSchedule): Punis
     consecutiveChargeNumber,
     consecutiveReportAvailable,
     damagesOwedAmount,
+    hasChildUnder18,
     canRemove,
     canEdit,
     rehabilitativeActivities,
@@ -323,11 +332,22 @@ export function flattenPunishment(punishment: PunishmentDataWithSchedule): Punis
     ...(consecutiveChargeNumber && { consecutiveChargeNumber }),
     ...(consecutiveReportAvailable && { consecutiveReportAvailable }),
     ...(damagesOwedAmount && { damagesOwedAmount }),
+    ...(hasChildUnder18 !== undefined && { hasChildUnder18 }),
     rehabilitativeActivities,
     rehabilitativeActivitiesCompleted,
     rehabilitativeActivitiesNotCompletedOutcome,
     previousSuspendedUntilDate,
   }
+}
+
+export function isSocialVisitsPunishment(type?: PunishmentType): boolean {
+  return [PunishmentType.RESTRICTION_OF_SOCIAL_VISITS, PunishmentType.LOSS_OF_SOCIAL_VISITS].includes(type)
+}
+
+export function parseHasChildUnder18(value: unknown): boolean | undefined {
+  if (value === true || value === 'true') return true
+  if (value === false || value === 'false') return false
+  return undefined
 }
 
 export function flattenPunishments(punishments: PunishmentDataWithSchedule[]): PunishmentData[] {
--- a/server/routes/punishmentsForReport/punishment/punishmentPage.ts
+++ b/server/routes/punishmentsForReport/punishment/punishmentPage.ts
@@ -11,6 +11,8 @@ import {
   PunishmentDataWithSchedule,
   PunishmentType,
   RehabilitativeActivity,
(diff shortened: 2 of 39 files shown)


server/services/punishmentsService.ts
```diff
@@ -438,11 +438,16 @@ export default class PunishmentsService {
     return new ManageAdjudicationsUserTokensClient(user).removePunishmentComment(chargeNumber, id)
   }
 
-  async checkAdditionalDaysAvailability(chargeNumber: string, user: User): Promise<boolean> {
+  async getPunishmentAvailability(
```
xoen, line 441: I like the idea of extracting the logic to determine which punishments are available in its own method!

However, the return type/value may be not communicate this intent.
I played with the idea of having an "available punishments" mapping from `PunishmentType` to `boolean`, e.g. _something like_:

```TypeScript
    const availablePunishments = {
      [PunishmentType.ADDITIONAL_DAYS]: isIndependentAdjudicatorHearing,
      [PunishmentType.PROSPECTIVE_DAYS]: isIndependentAdjudicatorHearing,
      [PunishmentType.CAUTION]: !(punishmentsAlreadyAdded || cautionAlreadyAdded),
      [PunishmentType.DAMAGES_OWED]: !damagesAlreadyAdded,
      [PunishmentType.LOSS_OF_SOCIAL_VISITS]: !reportedAdjudication.isYouthOffender,
      [PunishmentType.RESTRICTION_OF_SOCIAL_VISITS]:!reportedAdjudication.isYouthOffender,
    }
```

and then the template would use this information to decide whether a given punishment type is shown/not shown.

But don't worry about this change, I may have to touch this part as part of IR-1786, so I may well make these tweaks as part of that 👍

server/data/PunishmentResult.ts
```diff
@@ -323,13 +332,24 @@ export function flattenPunishment(punishment: PunishmentDataWithSchedule): Punis
...
     rehabilitativeActivitiesNotCompletedOutcome,
     previousSuspendedUntilDate,
   }
 }
 
+export function isSocialVisitsPunishment(type?: PunishmentType): boolean {
+  return [PunishmentType.RESTRICTION_OF_SOCIAL_VISITS, PunishmentType.LOSS_OF_SOCIAL_VISITS].includes(type)
+}
+
+export function parseHasChildUnder18(value: unknown): boolean | undefined {
```
xoen, line 347: is `value` `unknown` because we want to use this function on different inputs? e.g. the raw string but also the parsed boolean? Or is it `unknown` just in case? (if we are only interested in strings or booleans it could be `value: string | boolean` but I don't know if this is in place because of TS being difficult)

Also, (but I may missing something because I haven't looked at the whole PR diff), wouldn't be better if past this point the returned value would just be `boolean` rather than retain the ambiguity? e.g. couldn't we return false if value is not `true` nor `'true'`?

Potentially, couldn't this just be:

```TypeScript
export function parseHasChildUnder18(value: unknown): boolean {
  return value === true || value === 'true'
}
```

server/routes/punishmentsForReport/punishment/punishmentPage.ts
```diff
@@ -101,7 +114,18 @@ export default class PunishmentPage {
...
+      punishmentType,
+      privilegeType,
+      otherPrivilege,
+      stoppagePercentage,
+      damagesOwedAmount,
+      lossHasChildUnder18,
+      restrictionHasChildUnder18,
+    } = req.body
+    const hasChildUnder18 = parseHasChildUnder18(
+      punishmentType === PunishmentType.LOSS_OF_SOCIAL_VISITS ? lossHasChildUnder18 : restrictionHasChildUnder18,
```
xoen, line 127: Doesn't the child question apply to both social visits punishment types? 🤔  (sorry maybe I got that wrong and it only applies to _**loss**_ of social visits)

xoen: Scrap that, the `else`/`:` covers the restriction of social visits punishment.

server/data/PunishmentResult.ts
xoen: OK, so the reason why `value: unknown` is because it comes from `req.body` and it could be `string` or missing/`undefined`, and possibly other types.

Which presumably it's the reason why it's `unknown` and the return type could be `undefined`.

server/views/pages/autoPunishmentSchedule.njk
```diff
@@ -28,6 +28,9 @@
         <input type="hidden" name="_csrf" value="{{ csrfToken }}"/>
 
         <h2 class="govuk-heading-m" data-qa="punishment-name">{{ type | convertPunishmentType(stoppagePercentage, privilegeType, otherPrivilege) }}</h2>
+        {% if type === PunishmentType.LOSS_OF_SOCIAL_VISITS or type === PunishmentType.RESTRICTION_OF_SOCIAL_VISITS %}
```
xoen: I wonder if we could expose the `isSocialVisitsPunishment()` function as a Nunjucks filter and reuse it here, e.g.

```Nunjucks
{% if type | isSocialVisitsPunishment %}
   <!-- [...] --> 
```

it would read better and you remove the duplication/risk of this going out of sync with that logic (if let's say a new social visits punishment is introduced in the future...although I do appreciate that's YAGNI)

eve-bramley (approved):
LGTM

server/data/PunishmentResult.ts
cugorji-justice: Yes

server/views/pages/autoPunishmentSchedule.njk
cugorji-justice: Good point, I have made the recommended modifications

server/routes/punishmentsForReport/punishment/punishmentValidation.ts
```diff
@@ -64,6 +70,16 @@ export default function validateForm({
...
   }
 
+  if (isSocialVisitsPunishment(punishmentType) && hasChildUnder18 === undefined) {
+    return {
+      ...errors.MISSING_CHILD_UNDER_18,
+      href:
+        punishmentType === PunishmentType.LOSS_OF_SOCIAL_VISITS
+          ? '#lossHasChildUnder18'
+          : '#restrictionHasChildUnder18',
+    }
```
xoen: Maybe two entries should be added to the `errors` mapping for consistency, e.g.:

```TypeScript
MISSING_CHILD_UNDER_18_LOSS_OF_SOCIAL_VISITS: {
  href: '#lossHasChildUnder18',
  text: 'Select whether the prisoner has any children under 18',
},
MISSING_CHILD_UNDER_18_RESTRICTION_OF_SOCIAL_VISITS: {
  href: '#restrictionHasChildUnder18',
  text: 'Select whether the prisoner has any children under 18',
},
```

granted, it's a bit ugly and duplicated 🤔 

For example, I am now realising that the `href` in the error above was just a placeholder and there is no HTML input with that id right? (`#hasChildUnder18`)

Having said that, this is _not wrong_, so it's just an idea ☺️ .

xoen (approved):
Really nice work Chux!!

server/routes/punishmentsForReport/punishment/punishmentValidation.ts
cugorji-justice: This is done

xoen (approved):

PR #1467, opened 2026-07-21
IR-1785 Restrict prisoner adjudication data to the user's own caseloads

Mjwillis:
## Problem

A user could change the prisoner number in the URL to view the adjudication history, active punishments and individual charges of any offender in the country.

This cannot be fixed in the API. The frontend calls the adjudications API with a **system token** and tells it which caseload to act as via the `Active-Caseload` header, so the API's own check (`ReportedAdjudicationBaseService.findByChargeNumber`) is only as trustworthy as the header we send it. Several prisoner-scoped endpoints (`/booking/{bookingId}`, `/bookings/prisoner/{prisonerNumber}`, `/punishments/{bookingId}/active`) are deliberately cross-agency and do no caseload check at all.

Four gaps:

| Route | Problem |
|---|---|
| `activePunishments` | No check at all — any prisoner's live punishments, nationally |
| `adjudicationConsolidatedView` | The user-supplied `?agency=` query param was passed straight through as the API's `Active-Caseload` header, so the URL author chose which caseload the API validated against |
| `prisonerRoutes` (`/prisoner/:prisonerNumber/image`) | Unguarded prisoner photo |
| `adjudicationHistory` + consolidated view | Had a check from LSM-25, but fetched all the data first and only hid it at render time |

## Changes

- New `prisonerIsInUsersCaseloads` (`server/utils/caseloadHelper.ts`), checking the prisoner's agency against **every** caseload on the user's account rather than only their currently active one.
- All four routes guarded, and the check now runs **before** any data is fetched rather than hiding it in the template.
- The `?agency=` param is retained (it is needed to open a prisoner's historic charges from other establishments) but no longer reaches the API until we know the prisoner is in the user's caseloads *and* the charge belongs to that prisoner.
- Shared "you do not have permission" copy extracted into `server/views/partials/forbiddenEstablishment.njk`.

## ⚠️ Behaviour change — needs sign-off

**Access is no longer granted by the `GLOBAL_SEARCH` role.** Previously anyone with it could view adjudication records for any prisoner nationally. This will affect regional/HQ users, and anyone looking at a prisoner who has transferred to them but whose move has not yet landed in NOMIS.

`GLOBAL_SEARCH` is left untouched on the separate "search for a prisoner to transfer in" flow (`prisonerSearch.ts`).

## Testing

- Full unit suite passes (1172), including 9 new route tests and a 5-case helper suite.
- The new denial tests were confirmed to bite: stubbing the helper to return `true` makes exactly those 4 tests fail.
- 26 Cypress tests across the three affected pages pass.
- Lint and typecheck clean.

Not tested against real NOMIS data, so the `GLOBAL_SEARCH` change has not been exercised with real users' caseloads.

## How to verify manually

1. A prisoner at your active caseload — history, active punishments and a charge all load as before.
2. A prisoner at another caseload on your account, **without** switching active caseload — should also load.
3. A prisoner at a prison not on your account, for each of `/adjudication-history/:prisonerNumber`, `/active-punishments/:prisonerNumber` and `/prisoner-report-consolidated/:prisonerNumber/report/:chargeNumber` — permission message, no prisoner data anywhere in the response body.
4. Append `?agency=<other prison>` to the consolidated view URL for that prisoner — still refused.

### DIFF
--- a/server/routes/activePunishments/activePunishments.test.ts
+++ b/server/routes/activePunishments/activePunishments.test.ts
@@ -59,4 +59,42 @@ describe('GET /active-punishments', () => {
         expect(response.text).toContain('James Smith’s active punishments')
       })
   })
+
+  it('should load the details for a prisoner in a non-active caseload the user holds', () => {
+    reportedAdjudicationsService.getPrisonerDetails.mockResolvedValue(
+      testData.prisonerResultSummary({
+        offenderNo: 'G7234VB',
+        firstName: 'James',
+        lastName: 'Smith',
+        agencyId: 'LEI',
+      }),
+    )
+
+    return request(app)
+      .get(adjudicationUrls.activePunishments.urls.start('G7234VB'))
+      .expect('Content-Type', /html/)
+      .expect(response => {
+        expect(response.text).toContain('James Smith’s active punishments')
+      })
+  })
+
+  it('should not show punishments for a prisoner outside the users caseloads', () => {
+    reportedAdjudicationsService.getPrisonerDetails.mockResolvedValue(
+      testData.prisonerResultSummary({
+        offenderNo: 'G7234VB',
+        firstName: 'James',
+        lastName: 'Smith',
+        agencyId: 'BXI',
+      }),
+    )
+
+    return request(app)
+      .get(adjudicationUrls.activePunishments.urls.start('G7234VB'))
+      .expect('Content-Type', /html/)
+      .expect(response => {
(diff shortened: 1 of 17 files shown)


Mjwillis:
Superseded by #1474, which enforces the same restriction using the shared `hmpps-prison-permissions-lib` (the permission logic the Prisoner Profile uses) instead of a bespoke check. Closing this in favour of that approach.

PR #1473, opened 2026-07-24
IR-1786: Don't show 'Extra work'/'Removal from activity' for adults

xoen:
Policy has confirmed that the following punishments don't apply to adult:
- Extra work
- Removal from activity

This PR removes these from the list of punishments when Rule 55 applies (Adults).

Example of list of punishments for an adult, notice how the two above are not shown:

Screenshot 2026-07-24 at 15 32 44

**Other changes**:
- renamed `forfeitureOfVisitsDaysMax` to `lossOfVisitsDaysMax` in `AdjudicationResultReportDataBlank` class
  - we are moving away from the use "forfeiture" in favour of "loss" as it's an easier word to understand/spell/pronounce
- tweaked ESlint config to allow names exports
  - default exports can accidentally promote inconsistency and confusion, e.g. same thing imported using different names
- added mapping with punishments availability instead of passing (growing) number of variables to the template/macro
- refactored `punishmentRadios()` Nunjucks macro to:
  - use above mapping argument instead of several ones
  - add options to the list in the same order as they're shown on page, e.g. always use `push()` instead of a confusing mix of `push()`/`unshift()`
- use positive names and avoid double negation when possible (e.g. `if not cautionUnavailable ...`)


jonesj2193 (approved):
From what I can tell it looks good

PR #1474, opened 2026-07-24
IR-1785 Restrict adjudication access using shared prison-permissions library

Mjwillis:
## Why

A user could read any prisoner's adjudication data by hand-editing the URL. The frontend calls the adjudications API with a system token and tells it which caseload to act as via the `Active-Caseload` header, so this can only be enforced in the frontend.

Rather than a bespoke check, this adopts the shared **`@ministryofjustice/hmpps-prison-permissions-lib`** — the same library the Prisoner Profile and DPS micro-frontend components use — so the rules are consistent everywhere: the link to an adjudication on the profile and a hand-typed adjudications URL now obey identical rules.

> Supersedes #1467 (the earlier bespoke, caseloads-only approach). That PR can be closed.

## Permission used

Gates on `PrisonerAdjudicationsPermission.read` — the adjudications-specific permission the profile uses. Its rule:

- Prisoner in **any** of the user's caseloads → allowed
- **`GLOBAL_SEARCH` alone does not grant access** (consistent with the IR-1785 intent to drop it)
- **POM / Reception** users get cross-caseload access, plus restricted-patient / transferring / released handling — the nationally-agreed rule the profile already applies

## What changed

- `PermissionsService` constructed in the service registry (reuses the existing system-token client — no new auth dependency)
- Middleware maps `res.locals.user` onto the shape the library reads (`authSource`, `caseLoads`, `activeCaseLoadId`, `userRoles`)
- `prisonerPermissionsGuard` on adjudication history, active punishments and the consolidated report view
- Prisoner image route performs the equivalent check manually and serves the placeholder image on denial (it must not 403)
- Consolidated view keeps its charge-ownership check (the `?agency=` tamper defence the guard can't cover)
- Denials render a shared forbidden page via the error handler — which previously redirected **every** 403 to sign-out
- Removed the old `forbidden` / `GLOBAL_SEARCH` fetch-then-hide code and template branches

## ⚠️ Behaviour change

`GLOBAL_SEARCH` users lose out-of-caseload adjudication access; **POM and Reception** users gain cross-caseload access. This matches the Prisoner Profile.

## ⚠️ Deploy prerequisite

The adjudications **system client must hold `ROLE_VIEW_PRISONER_DATA`** (helm / cloud-platform change) so the library can read hmpps-prisoner-search. Without it, every request is denied.

## Testing

- Unit tests: grant/deny cases for all four surfaces + the user-shape adapter middleware (1173 passing)
- Lint and typecheck clean
- Integration (Cypress) specs updated with a prisoner-search stub for the guard — **please confirm the Cypress run in CI**
- Manual end-to-end against real prisoner-search still to be done once `ROLE_VIEW_PRISONER_DATA` is granted in the environment

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
   "dependencies": {
     "@ministryofjustice/frontend": "^9.0.0",
     "@ministryofjustice/hmpps-monitoring": "^2.1.0",
-    "@ministryofjustice/hmpps-prison-permissions-lib": "4.2.0",
+    "@ministryofjustice/hmpps-prison-permissions-lib": "4.1.0",
     "@ministryofjustice/hmpps-rest-client": "2.2.0",
     "@types/jsonwebtoken": "^9.0.10",
     "agentkeepalive": "^4.6.0",
--- a/package.json
+++ b/package.json
@@ -50,6 +50,8 @@
   "dependencies": {
     "@ministryofjustice/frontend": "^9.0.0",
     "@ministryofjustice/hmpps-monitoring": "^2.1.0",
+    "@ministryofjustice/hmpps-prison-permissions-lib": "4.2.0",
+    "@ministryofjustice/hmpps-rest-client": "2.2.0",
     "@types/jsonwebtoken": "^9.0.10",
     "agentkeepalive": "^4.6.0",
     "applicationinsights": "^3.15.0",
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -80,6 +80,7 @@ export default defineConfig({
         stubGetUserFromUsername: users.stubGetUserFromUsername,
         stubSearch: prisonerSearch.stubSearch,
         stubSearchPrisonerDetails: prisonerSearch.stubSearchPrisonerDetails,
+        stubGetPrisonerPermissionDetails: prisonerSearch.stubGetPrisonerPermissionDetails,
         stubAuthUser: users.stubAuthUser,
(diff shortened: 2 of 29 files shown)


Mjwillis:
Update on the deploy prerequisite: the system client (`hmpps-manage-adjudications-ui-client`) already holds `ROLE_VIEW_PRISONER_DATA`, so no environments/helm change is needed — the library's prisoner-search calls will authenticate as expected. Remaining pre-merge check is just the Cypress run in CI.

package.json
```diff
@@ -50,6 +50,8 @@
   "dependencies": {
     "@ministryofjustice/frontend": "^9.0.0",
     "@ministryofjustice/hmpps-monitoring": "^2.1.0",
+    "@ministryofjustice/hmpps-prison-permissions-lib": "4.2.0",
```
theDustRoom: Could we drop `@ministryofjustice/hmpps-prison-permissions-lib` to `4.1.0` here?

This would avoid a timing issue with the new template-aligned `.npmrc` `min-release-age = 7` setting, as `4.2.0` is currently only about 3 days old and may be blocked by future installs until it ages past 7 days.

theDustRoom (changes requested):
We have another PR that aligns `.npmrc` with the TypeScript Template, this will cause an issue with the `@ministryofjustice/hmpps-prison-permissions-lib` dependency at `4.2.0` as it's only 3 days old.

theDustRoom (approved):

